PR #1338, opened 2013-03-30
adds settings menu to ace

matthewkastor:
I'm not stuck on the keybinding (control-q) or aesthetics of this, please change
them at will. All I really want is the ability to display a menu where I
can change the editor's settings from and, I want the menu to be smart
enough to automatically know about new settings. This pretty much works
except there isn't a registry for themes and modes so every new theme or
mode has to be added to the internal list by hand...


LeviRosol:
Are there any options to limit the items that appear in this menu? As a simple example, I want to remove `mode` and `keyboard handler` from the menu.

PR #1599, opened 2013-09-10
Use native HTML5 Drag'n'Drop for text.

danyaPostfactum:
Sorry, I donno how to rebase branch, so this is the new version of https://github.com/ajaxorg/ace/pull/1431 .
All issues I wanted are fixed.

This needs test and review.

Issues I would like to fix but need help:
- [ ] Wrong selection on undo of drag operation.
- [x] Autoscroll while dragging should not scroll if there is no line (scrolled past last row) or char (scrolled past last char) at cursor.
- [x] Editor should not scroll selection into view if we leave (through negligence) editor rect while autoscrolling.

Issues we can not resolve:
- Wrong ui feedback (drag cursor) in IE9+
- Safari 5 does not support dnd

Issues I do not know reason of:
- Opera blinks drag cursor (cursor icon) while dragging.

Issues I don't like or not sure:
- Sequence click interval is handled by browsers differently.

Also, in Firefox, dragenter event fires twice, so dragleave event does not get caret back. Can anybody reproduce?

Great news! All bugs I (or someone else) found in IE and reported to MS are marked as Won't Fix )))
They say those are "by design", "not reproducible" etc :)
1) https://connect.microsoft.com/IE/feedback/details/789773/ie-increments-event-detail-on-clicks-over-different-locations
2) https://connect.microsoft.com/IE/feedback/details/790344/ie-ignores-draggable-attribute-if-set-after-mousedown-event
3) https://connect.microsoft.com/IE/feedback/details/782906/setting-dropeffect-has-no-effect-in-ie9-and-ie10-standards-mode
4) https://connect.microsoft.com/IE/feedback/details/780585/dropeffect-not-working-properly-in-ie-standard-mode-ie9-ie10

### DIFF
--- a/lib/ace/mouse/default_handlers.js
+++ b/lib/ace/mouse/default_handlers.js
@@ -47,7 +47,7 @@ function DefaultHandlers(mouseHandler) {
     editor.setDefaultHandler("quadclick", this.onQuadClick.bind(mouseHandler));
     editor.setDefaultHandler("mousewheel", this.onMouseWheel.bind(mouseHandler));
 
-    var exports = ["select", "startSelect", "selectEnd", "selectByWordsEnd",
+    var exports = ["select", "startSelect", "selectEnd", "selectAllEnd", "selectByWordsEnd",
         "selectByLinesEnd", "dragWait", "dragWaitEnd", "focusWait"];
 
     exports.forEach(function(x) {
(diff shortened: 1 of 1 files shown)


mlajtos:
Why [this branch](https://rawgithub.com/danyaPostfactum/ace/dragimage/kitchen-sink.html) was not merged? It gives great feeback while dragging the selected text.

PR #1745, opened 2014-01-01
Refactor document delta handling

aldendaniels:
*The changes proposed below seek to simplify Ace's handling of document mutations in the interest of bug-free code and easy integration. I hope that the benefits of the recommended changes will be clear, but I'd be glad to elaborate as necessary.*

*Feedback and review of these changes from a core contributor would be appreciated. Having extensive experience with web-based editing, I realize the risk of destabilization inherent in touching the core document code. In this case, I think it's worth the trade-off, but the changes will need careful scrutiny.*

**Refactor delta types**

- Combine the ```insertText``` and ```insertLines``` delta types into a single ```insert``` delta type

- Combine the ```removeText``` and ```removeLines``` delta types into a single ```remove``` delta type

The new delta types will store inserted or deleted text has an array of lines (split on the detected new line character). This is efficient and has no presuppositions regarding newlines. Pressing ENTER in an empty document, therefore, will create the following delta:
```javascript
{
    "action": "insert",
    "range": {
        "start": {
            "row": 0,
            "column": 0
        },
        "end": {
            "row": 1,
            "column": 0
        }
    },
    "lines": ["", ""]
}
```

**Handle all document changes in a single spot**

- Make all document mutations in a single applyDelta function.

- Add delta validation

Basically ```insert()```, ```insertLines()```, ```insertNewLine()```, ```remove()```, etc. should all wrap the ```applyDelta()``` function instead of the inverse. This way it's easy to ensure that all deltas are being applied in the same way and are completely reversible. This also makes it easy to validate deltas in a single location.

**Supporting changes**

- Rework / Simplify anchor logic to handle new delt types

- Rename:
    - ```insert()``` to ```insertText()``` 
    - ```remove()``` to ```removeText()```
    - ```insertLines()``` to ```insertFullLines()```
    - ```removeLines()``` to ```removeFullLines()```

### DIFF
--- a/lib/ace/anchor.js
+++ b/lib/ace/anchor.js
@@ -133,7 +133,7 @@ var Anchor = exports.Anchor = function(doc, row, column) {
             }
             
             // DELTA ENVELOPS POINT (delete only): Move point to delta start.
-            if (delta.action != 'delete')
+            if (delta.action != 'remove')
                 throw 'Delete action expected.';
             
             return {
--- a/lib/ace/anchor.js
+++ b/lib/ace/anchor.js
@@ -100,50 +100,50 @@ var Anchor = exports.Anchor = function(doc, row, column) {
      *
      **/
     this.onChange = function(e) {
+        var delta = e.data;
+        var point = this.$getTransformedPoint(delta, {row: this.row, column: this.column});
+        this.setPosition(point.row, point.column, true);
+    };
+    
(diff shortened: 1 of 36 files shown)


lennartcl:
Closing this in favor of https://github.com/ajaxorg/ace/pull/1819. Also note the link in the changelog there.

PR #1850, opened 2014-03-12
Key simulation

rksm:
This implements a simple form of triggering commands and insertions by simulating keys. Like:

``` js
var keys = require('ace/lib/keys');
// one key combo at a time
keys.simulateKey(editor, 'Shift-Right');
keys.simulateKey(editor, 'Command-Shift-d');
keys.simulateKey(editor, 'Control-a');
// or multiple
keys.simulateKeys(editor, 'H e l o Left l Command-Shift-Left');
/*...*/
```

todo:
- [x] support series of keys (like `keys.simulateKeys(editor, 'Right Right Command-Down')`)
- [x] base `keys.send` / old vim sendKeys code upon simulateKeys

### DIFF
--- a/lib/ace/lib/keys.js
+++ b/lib/ace/lib/keys.js
@@ -180,7 +180,7 @@ exports.getModifierString = function(e) {
  * @return {Boolean} return value that'll be returned by the keydown/keypress event handler
  *
  **/
-exports.normalizeCommandKeys = function(callback, e, keyCode, pressedKeys) {
+exports.normalizeCommandKeys = function normalizeCommandKeys(callback, e, keyCode, pressedKeys) {
     var hashId = getModifierHash(e);
 
     if (!useragent.isMac && pressedKeys) {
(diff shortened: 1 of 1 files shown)


knod:
Has the ability to simulate key presses been added in some fashion? Not just content keys, but modifier keys as well and the tab key functionality.

To take tab as an example - just inserting a tab works for most occasions, and indenting might be handled as described here: http://stackoverflow.com/questions/24607730/ace-editor-javascript-triggering-a-tab-press-event-for-ace-editors-event-hand/24608974#24608974. Unfortunately, more uses of tab keep popping up. With snippetManager, pressing tab navigates to the next specified text in a snippet and I want that functionality as well.

I don't want to have to recreate all the functionality for each plugin I use. I'd like to just simulate an actual key being pressed or held. Is this possible?

rksm:
@knod Since this feature seems to be unwanted I ended up publishing it in an [extension](https://github.com/rksm/ace.improved/blob/master/lib/ace.ext.keys.js). `ace.ext.keys.simulateKeys()` works with [modifiers and key sequences](https://github.com/rksm/ace.improved/blob/master/tests/ace.ext.keys-test.js).

rksm:
@knod ace.improved has some doc now, https://github.com/rksm/ace.improved#aceextkeys

knod:
@rksm Thanks, I'll look into that. I have a lot else to work on, but this will become a major issue soon, so I'm glad to see there's a way to handle it out there.

jakec-github:
Did this functionality ever get added? Could really use the ability to control the Vim mode programatically in a project I am working on

PR #1908, opened 2014-04-14
Add python worker (powered by skulpt)

danyaPostfactum:
I am not interested in python, but it taked 5 minutes to wrap a Skulpt into the worker.
Probably someone will like it.

Skulpt implements Python 2.x, so I guess we need the way to specify a version to disable worker for 3.x .
I guess it is possible to switch parser grammar tables to support both versions (not sure about AST).

It would be good to make a custom build of Skulpt to get rid of runtime module.

There are also error-reporting issues.


sheeeeep:
any progress with this pr ?

PR #2199, opened 2014-10-20
add liveAutocompletionAutoSelect option

adamjimenez:
When enabled this will select the first matching item when using live autocomplete.


adamjimenez:
Can't remember where I was going with this exactly, but I no longer need it.
Setting editor.completer.autoSelect seems to be sufficient.

PR #2275, opened 2014-12-15
change variable char to character

mehtaphysical:
char is a reserved word in js. This was causing problems for me when trying to minify ace editor with yui compress. I believe it is best to change this variable name.


Huawei8 (commented):
Update

PR #2352, opened 2015-02-11
Fix autocomplete prefix for custom identifierRegexes

dmarcotte:
Autocomplete prefix can be wrong for completions with a custom identifierRegex because `Autocompleter.base` is computed relative to a prefix defined by the default identifierRegex.

Steps to reproduce the issue this pull request solves:
- create a completer whose completions provide more permissive `identifierRegex` than the default (say one that matches `$` and `{` in addition to the defaults)
- trigger auto-completes after typing `${`, and note how the prefix is respected:
    screen shot 2015-02-10 at 1 35 15 pm

- continue typing, and note how the `${` part of the prefix is lost and we only match against `test`:
    screen shot 2015-02-10 at 1 36 10 pm

With this patch, we instead get the expected prefix of `${test`:

screen shot 2015-02-10 at 2 46 17 pm

### DIFF
--- a/lib/ace/autocomplete.js
+++ b/lib/ace/autocomplete.js
@@ -220,27 +220,34 @@ var Autocomplete = function() {
         "PageDown": function(editor) { editor.completer.popup.gotoPageDown(); }
     };
 
-    this.gatherCompletions = function(editor, callback) {
-        var session = editor.getSession();
-        var pos = editor.getCursorPosition();
-
+    /**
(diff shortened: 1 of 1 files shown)


dmarcotte:
This has since been fixed in https://github.com/ajaxorg/ace/commit/fcebd0f662e0434fadf15faaaf4b59ab8950217a.  Thanks!

PR #2538, opened 2015-06-09
Adds hasOwnProperty check for cloned Objects

alvincrespo:
This is necessary for applications where Object.prototype is extended. In our case we use Ember with native prototype extensions enabled. Without this check, Ember's extensions are copied over and error out in the app.


alvincrespo:
Ah. NVM - this works with the patch here: https://github.com/ajaxorg/ace/pull/2435. Looks like ace-builds hasn't created a tag so I was getting 1.1.9 and now what was on master.

PR #2647, opened 2015-08-21
Praat

jjatria:
Commands that follow "local" variables (variables with dots in their names) were not correctly identified. Editor-mode commands were also not properly recognised when following directives.

This pull request fixes both of those problems, and adds some test conditions to the demo file.


jjatria:
Wow, I had completely forgot about this!
Thanks for following up on it after all this time!

PR #2899, opened 2016-02-17
[FIX] end tag with space after name should be valid

gdcGithub:
xml End tag may have spaces after name.
<a></a   >
is a valid xml.

### DIFF
--- a/lib/ace/mode/xml/sax.js
+++ b/lib/ace/mode/xml/sax.js
@@ -98,7 +98,10 @@ function parse(source,defaultNSMapCopy,entityMap,domBuilder,errorHandler){
 		}
 		switch(source.charAt(i+1)){
 		case '/':
-			var end = source.indexOf('>',i+3);
+			var regex = new RegExp("\\s*>");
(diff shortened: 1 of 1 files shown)


lib/ace/mode/xml/sax.js
```diff
@@ -448,10 +451,17 @@ function fixSelfClosed(source,elStartEnd,tagName,closeMap){
...
 	if(pos == null){
 		//console.log(tagName)
-		pos = closeMap[tagName] = source.lastIndexOf('</'+tagName+'>')
+		var regex = new RegExp("</"+tagName+"\\s*>", "gi");
+		var match      = source.match(regex);
+		if (match && match.length >= 1) {
+			var lastIndex  = source.lastIndexOf(match[match.length-1]);
+			pos = closeMap[tagName] = lastIndex;
+		} else {
+			pos = -1;
```
marinsokol5, line 460: Can we also keep the same chain assignment from above here, so `pos = closeMap[tagName] = -1` just to keep the  previous logic the same.

PR #2931, opened 2016-03-25
Optimize for loops

Noctisdark:
Changed ```for(var i = something; i < otherStuff.length; i++) ;``` to ```for(var i = something; l = otherStuff.length; i < l; i++);```. It turns out that this is faster, At least stackoverflow agree http://stackoverflow.com/questions/1340589/are-loops-really-faster-in-reverse. https://blogs.oracle.com/greimer/resource/loop-test.html
I know this won't make ace 2 times faster or so, but I think it worth to try it out!


Noctisdark:
Hmmm, I guess missed something ^^"

PR #2957, opened 2016-04-28
Adjusted "insert" so minification couldn't wrongly replace whitespace.

GlitchHound:
Found while building an app that minification ate "\t" on line 974 and replaced it with " ". This is one, probably naive, way to prevent that happening again.

Symptom of the problem: even with useSoftTabs disabled, spaces entered were being replaced with a number of spaces equal to session.$tabSize

Symptom existed only when using minified sources, pulled both from ajaxorg/ace-builds and from the jsdelivr CDN. In chrome, the resultant error is in the "pretty printed" minified file at line 9730.

### DIFF
--- a/lib/ace/editor.js
+++ b/lib/ace/editor.js
@@ -971,7 +971,7 @@ var Editor = function(renderer, session) {
             }
         }
         
-        if (text == "\t")
+        if (text == String.fromCharCode(9))
             text = this.session.getTabString();
 
         // remove selected text


GlitchHound:
In the shower it occurred to me that this probably wouldn't have gone unnoticed and there must be some other explanation.

Turns out there was. Somewhere along the line in my infinite wisdom I decided to copy one of these files, paste it, and save that. Downloading them doesn't have the missing tab issue.

TL;DR: Don't copy and paste code, kids.

Issue #3037, opened 2016-07-25
Unicode Characters ⟦ and ⟧ Create Spacing/Cursor Position Problem


Dmitri-2:
This issue occurs when using the ⟦ and ⟧ Unicode characters in the Ace editor. (U+27E6 and U+27E7, respectively.) Using these characters within the editor causes the cursor position to be incorrect, likely due to the fact that these characters have a smaller space consumption than the corresponding regular double bracket (compare [[ and ⟦). The more of these characters on one line, the worse the spacing problem is. Screenshots below: 

![screen shot 2016-07-25 at 2 56 05 pm](https://cloud.githubusercontent.com/assets/19738862/17118962/0c02d938-5278-11e6-810b-a28b051cb7a0.png)

and  

![screen shot 2016-07-25 at 2 56 34 pm](https://cloud.githubusercontent.com/assets/19738862/17118978/205b9dac-5278-11e6-944b-a34ab4620781.png)

Dmitri-2:
@12Me21 We came up with a specialized fix for our purposes, by specifically checking for the bracket characters in `lib/ace/edit_session.js`. This allows us to manually set the screen width of the character and avoid any further errors. (See lines 2000-2037 of edit_session.js). It would be great if Ace was updated to provide full unicode support though! Also, for more info on character support see the `varchar` branch of the Ace editor.

PR #3137, opened 2016-11-08
Support for Arabic/Hebrew/Persian languages.

ashensis:
This PR is intended to provide ACE with proper Arabic/Hebrew/Farsi (aka. Bidi) characters typing/display. (references the discussion under issue 'RTL editing' #1429)
Currently ACE fails to do this for 2 primary reasons.
1. ACE bases itself on monospaced font in calculating screen columns offsets.
Unfortunately, Bidi character's width differs from that of regular (say English)
characters. This poses challenge, especially when Bidi and none Bidi (English or neutral) characters
are present in the same line.
Moreover, for Chrome engine based browsers (like Chrome and Opera), on some OS the default 'Monako' font
defined in editor.css is not monospaced when it comes to Bidi characters. 
This might happen or not happen even on different Windows7 systems.
Apparently this happens since 'Monaco' has no glifs for Bidi characters and font substitution
on different OS(s) picks up different fonts from awailable stock.

2. Bidi languages differ from none Bidi ones in their flow direction (from right to left),
thus, character order on display (right-to-left) differs from respective order in text buffer (left-to-right).
The respective character positions on display versus buffer are defined by Bidi algorithm implemented by platform
(either by OS for native applications or by browser for web based applications).
The correspondence between respective character positions on display (aka Visual) and in buffer (aka Logical)
is defined by Visual<->Logical maps produced by Bidi Engine implementing Bidi algorithm.

3. Between additional miscellaneous challenges it worth mentioning:
- Arabic LamAlef pair processing (this particular combination of two characters is displayed as one character) 
- 'Tab' and 'wide' characters processing (their width participates in calculating the screen column position offset)

For references to Bidi algorithm see, for instance, http://unicode.org/reports/tr9/
For general references regarding Bidi see, for instance, https://www.w3.org/International/articlelist#direction 

As a consequence, the character screen position (offset) in pure Bidi or mixed Bidi/None-Bidi text lines
should be calculated basing on character screen position rather then buffer one.

The following implementation details are important for understanding how this PR deals with aforementioned problems.

1. Statefull 'bidihandler' class is introduced to concentrate all Bidi related activities.
It depends on lightweight Bidi engine implemented by library class 'bidiutil'.

2. 'bidihandler' class stores information related to text line under processing, this info 
consists of so named 'bidiMap' that contains Bidi levels and Visual<->Logical map (see above for terms definition).
Above mentioned information is updated when caret is moved to another text line.
 
3. In order to cope with the problem of calculating correct horizontal offsets of displayed characters 
'bidihandler' class stores width of Bidi characters alongside with widths for None Bidi characters, end-of-lines etc.
In cases when 'Monaco' font doesn't furnish monospaced glyphs for Bidi characters, this font gets dynamically 
replaced by "Courier' font on per line basis (for lines containing Bidi characters) 

4. Special methods of 'bidihandler' class, like 'getPosLeft', 'offsetToCol', 'getSelections' ensure correct caret display, 
calculation of mouse click position and text (single or multiple) selection.

This solution has been tested on IE, FF, Chrome/Opera, Safary.

The test cases will be added to this PR after current approach is approved and code implementation is finalized.

### DIFF
--- a/lib/ace/bidihandler.js
+++ b/lib/ace/bidihandler.js
@@ -61,6 +61,8 @@ var BidiHandler = function(session) {
     this.wrapIndent = 0;
     /* font face has been set to Courier to ensure monospaced Bidi characters in Chrome/Opera */
     this.isCourierFont = false;
+    this.isLastRow = false;
+    this.EOF = "\xB6";
 };
 
 (function() {
@@ -120,6 +122,7 @@ var BidiHandler = function(session) {
 			docRow = this.getDocumentRow();
 			
 		this.wrapIndent = 0;
+		this.isLastRow = (docRow === this.session.getLength() - 1);
 		this.line = this.session.getLine(docRow);
 		if (this.session.$useWrapMode) {
 			var splits = this.session.$wrapData[docRow];
@@ -150,8 +153,8 @@ var BidiHandler = function(session) {
 	};
 	
 	this.updateBidiMap = function() {
-		var textCharTypes = [];
-		var line = this.line + (this.showInvisibles ? this.EOL : bidiUtil.DOT);
+		var textCharTypes = [], endOfLine = this.isLastRow ? this.EOF : this.EOL;
+		var line = this.line + (this.showInvisibles ? endOfLine : bidiUtil.DOT);
 		if (bidiUtil.hasBidiCharacters(line, textCharTypes)) {
 			this.bidiMap = bidiUtil.doBidiReorder(line, textCharTypes, this.isRtlDir);
 		} else {
--- a/lib/ace/bidihandler.js
+++ b/lib/ace/bidihandler.js
@@ -1,0 +1,334 @@
+/* ***** BEGIN LICENSE BLOCK *****
+ * Distributed under the BSD license:
+ *
+ * Copyright (c) 2010, Ajax.org B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of Ajax.org B.V. nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL AJAX.ORG B.V. BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * ***** END LICENSE BLOCK ***** */
+
+define(function(require, exports, module) {
+"use strict";
+
+var bidiUtil = require("./lib/bidiutil");
+var lang = require("./lib/lang");
+
+/**
+ * This object is used to ensure Bi-Directional support (for languages with text flowing from right to left, like Arabic or Hebrew)
+ * including correct caret positioning, text selection mouse and keyboard arrows functioning
+ * @class BidiHandler
+ **/
+
+/**
+ * Creates a new `BidiHandler` object
+ * @param {EditSession} session The session to use
+ *
+ * @constructor
+ **/
+var BidiHandler = function(session) {
+    this.session = session;
+    this.bidiMap = {};
+    /* current screen row */
+    this.currentRow = null;
+    this.bidiUtil = bidiUtil;
+    /* Arabic/Hebrew character width differs from regular character width */
+    this.charWidths = [];
+    this.EOL;
+    this.showInvisibles;
+    this.isRtlDir = false;
+    this.line = "";
+    this.wrapIndent = 0;
+    /* font face has been set to Courier to ensure monospaced Bidi characters in Chrome/Opera */
+    this.isCourierFont = false;
+};
+
+(function() {
+    /**
+     * Returns 'true' if row contains Bidi characters, in such case
+     * creates Bidi map to be used in operations related to selection
+     * (keyboard arrays, mouse click, select)
+     * @param {Number} the screen row to be checked
+     * @param {Number} the document row to be checked [optional]
+     * @param {Number} the wrapped screen line index [ optional]
+     **/
+	this.isBidiRow = function(screenRow, docRow, splitIndex) {
+		if (screenRow !== this.currentRow) {
+			this.currentRow = screenRow;
+			this.updateRowLine(docRow, splitIndex);
+			this.updateBidiMap();
+		}
+		return this.bidiMap.bidiLevels;
+	};
+
+	this.hasBidiCharacters = function(str) {
+		return /[\u0591-\u06ff\ufb1d-\ufefc]/.test(str);
+	};
+
+	this.getDocumentRow = function() {
+		var docRow = 0;
+		var rowCache = this.session.$screenRowCache;
+		if (rowCache.length) {
+			var index = this.session.$getRowCacheIndex(rowCache, this.currentRow);
+			if (index >= 0)
+				docRow = this.session.$docRowCache[index];
+		}
+
+		return docRow;
+	};
+
+	this.getSplitIndex = function() {
+		var splitIndex = 0;
+		var rowCache = this.session.$screenRowCache;
+		if (rowCache.length) {
+			var currentIndex, prevIndex = this.session.$getRowCacheIndex(rowCache, this.currentRow);
+			while (this.currentRow - splitIndex > 0) {
+				currentIndex = this.session.$getRowCacheIndex(rowCache, this.currentRow - splitIndex - 1);
+				if (currentIndex !== prevIndex)
+					break;
+
+				prevIndex = currentIndex;
+				splitIndex++;
+			}
+		}
+
+		return splitIndex;
+	};
+
+	this.updateRowLine = function(docRow, splitIndex) {
+		if (docRow === undefined)
+			docRow = this.getDocumentRow();
+			
+		this.wrapIndent = 0;
+		this.line = this.session.getLine(docRow);
+		if (this.session.$useWrapMode) {
+		    var splits = this.session.$wrapData[docRow];
+		    if (splits) {
+		    	if (splitIndex === undefined)
+		    		splitIndex = this.getSplitIndex();
+
+		        if(splitIndex > 0 && splits.length) {
+		        	this.wrapIndent = splits.indent;
+		            this.line = (splitIndex < splits.length) ?
+		            	this.line.substring(splits[splitIndex - 1], splits[splits.length - 1]) :
+		            		this.line.substring(splits[splits.length - 1]);
+		        } else {
+		            this.line = this.line.substring(0, splits[splitIndex]);
+		        }
+		    }
+		}
+			
+		/* replace tab and wide characters by commensurate spaces */
+		var session = this.session, shift = 0, size;
+		this.line = this.line.replace(/\t|[\u1100-\uFFE6]/g, function(ch, i){
+			if (ch === '\t' || session.isFullWidth(ch.charCodeAt(0))) {
+				size = (ch === '\t') ? session.getScreenTabSize(i + shift) : 2;
+				shift += size - 1;
+				return lang.stringRepeat(bidiUtil.DOT, size);		
+			}
+		});
+	};
+	
+	this.updateBidiMap = function() {
+		var textCharTypes = [];
+		var line = this.line + (this.showInvisibles ? this.EOL : bidiUtil.DOT);
+		if (bidiUtil.hasBidiCharacters(line, textCharTypes)) {
+			this.bidiMap = bidiUtil.doBidiReorder(line, textCharTypes, this.isRtlDir);
+		} else {
+			this.bidiMap = {};
+		}
+	};
+
+    /**
+     * Resets stored info related to current screen row
+     **/
+	this.markAsDirty = function() {	
+		this.currentRow = null;
+	};
+
+    /**
+     * Updates array of character widths
+     * @param {boolena} is chrome based engine
+     * @param {Object} font metrics
+     *
+     **/
+	this.updateCharacterWidths = function(isChrome, $fontMetrics) {
+		var _regularCharWidth = $fontMetrics.$measureCharWidth('a'), _bidiCharWidth = $fontMetrics.$measureCharWidth('\u05d4');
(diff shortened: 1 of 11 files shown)


tomerm:
@ebraminio the fix for issues reported in #1429 was submitted in this PR. Do you know who should review it ? I just wonder if we need to attract anybody's attention to it.

ebraminio:
I am not familiar with Ace project maintainers but this looks good, even the fact I was not in favor of adding a bidi algorithm but if this in not avoidable, perhaps ace should go for it.

Also, I just spotted a minor issue on reverting functionality (double click a RTL word, remove it then press Ctrl+Z, highlighting area is misplaced) which IMO, this totally can go without it; so functionally, this LGTM to me.

BTW, "Pharsi" is misspelling of "Farsi" but even better [you can](https://en.wikipedia.org/wiki/Persian_language#Persian_language_name_in_English) use "Persian" to refer to the language.

Thanks for this great effort, I like to see this ASAP on websites using Ace.

tomerm:
@ebraminio thanks a lot for the feedback. This patch is far from being perfect / final. It is basically a prototype which illustrate basic approach to solution (and make essential features to work - cursor movement, typing etc.). Once we finalize the approach we will obviously polish the implementation to make sure all use cases are covered.

tomerm:
@nightwing it looks like you were involved in discussion and review of this issue from the very beginning. Would you please have a look at proposed solution and share your comments ? Many thanks in advance.

ashensis:
@ebraminio Thank you for your comments.
Regarding your remark about Ctrl+Z causing misplaced selection, I tried to reproduce this scenario with no success. Could you please specify the exact word you are selecting/deleting and what kind of browser & OS have been used?

ebraminio:
I am getting across different browsers consistently but I hope this would help to repro there, let me know
1. Open kitchen-sink.html, default setting
2. Put "متن متن متن متن" on it
3. Select second word, delete it with backspace
4. Ctrl+Z (Undo)
5. Ctrl+A (Select all)
image

Anyway, I think this has less importance for being a merge blocker. Thanks

ashensis:
Thank you very much for cooperation @ebraminio.
I already found the problem, now got it fixed.

rocketinventor:
@ashensis Thanks so much for the detailed write up. It was very informative.

nightwing:
Sorry for not finding time to review this earlier.
The changes here look fairly good, but there are still some issues that should be fixed to get this into mergeable state.

Hardcoding  "Courier" font doesn't work for mac, and doesn't work well with wrapping, since character limits are computed for the main font, maybe for now we should just ask users pick a monospace font that supports their language.

Caret movement is generally different from that of textarea. E.g. up/down keep logical column instead of the visual one, left/right move by logical column too, is this intentional?

ashensis:
Thank you @nightwing for your comments.
1. Regarding what you said about asking users pick a mono-spaced font that supports their language.
Let me make sense of this. Currently the list of preferred fonts is hard coded in editor.css as follows:
    font: 12px/normal 'Monaco', 'Menlo', 'Ubuntu Mono', 'Consolas', 'source-code-pro', monospace; 
So obviously it is a "administrator" choice to define a font to be used.
However the problem of aforementioned fonts that doesn't work for Arabic/Hebrew and few other languages is only relevant for "Chrome" engine browsers, like Chrome and Opera.
For IE or FF this problem doesn't exist.
In my code I automatically chose "Courier" only if lack of "monospacety" is detected and this should resolve above mentioned problem for "Chrome". Please note that in this case working with currently predefined fonts isn't an option at all from perspective of user experience..
I could add additional check for "mac" to avoid font switching on this OS. What do you think about this amendment?
2. You made a good point in observing that caret movement is different, as for up/down arrow keys it was my omission, this should be fixed to ensure "visual" movement across different lines. I will try to resolve this.
Regarding left/right arrow keys, this behavior of "logical" movement is standard across all editors in case of left-to-right base text direction, you can check this with Windows system editor (notepad) for instance, as well as with MS Word and any other Browser or OS based text editor (and ACE currently works with this left-to-right direction).
(It worth mentioning although, that in right-to-left base text direction the caret movement does coincide with arrow direction)

ashensis:
Fixed problem of up/down arrow keys to ensure "visual" movement across different lines.
Added check to exclude forcing "Courier" font on Chrome for 'MAC'

ebraminio:
Please rename "Pharsi" to "Persian" (or, "Farsi", which is [less favorable](https://en.wikipedia.org/wiki/Persian_language#Persian_language_name_in_English)) on the title.

lyudmil:
@nightwing Anything I can help with to get this merged in?

ashensis:
Thank you @nightwing.
Your testing is most welcome.
From my perspective the code is ready.

adir01:
@nightwing when you think our code can be merged ?

lib/ace/edit_session.js
```diff
@@ -1116,6 +1118,7 @@ var EditSession = function(text, mode) {
      *
      **/
     this.insert = function(position, text) {
+        this.$bidiHandler.markAsDirty();
```
ashensis: Frankly speaking I do not understand this proposal. 
Line 250 of edit-session.js falls with 'onChangeFold' method that has no bearing on our subject.
The intended changes of mine (invocation of 'markAsDirty') serve to rebuilt Bidi info pertinent to line being edited, in order to be able to correctly calculate caret position.

lib/ace/bidihandler.js
```diff
@@ -0,0 +1,338 @@
...
+(function() {
+    /**
+     * Returns 'true' if row contains Bidi characters, in such case
+     * creates Bidi map to be used in operations related to selection
+     * (keyboard arrays, mouse click, select)
+     * @param {Number} the screen row to be checked
+     * @param {Number} the document row to be checked [optional]
+     * @param {Number} the wrapped screen line index [ optional]
+     **/
+	this.isBidiRow = function(screenRow, docRow, splitIndex) {
```
ashensis: OK.

lib/ace/edit_session.js
```diff
@@ -2430,6 +2443,10 @@ var EditSession = function(text, mode) {
         this.$stopWorker();
     };
 
+    this.isFullWidth = function(c) {
+        return isFullWidth(c);
```
ashensis: OK.

lib/ace/layer/text.js
```diff
@@ -311,6 +319,21 @@ var Text = function(parentEl) {
             row++;
         }
         this.element.innerHTML = html.join("");
+        if (this.session.$bidiHandler.isCourierFontFace())
```
ashensis: If you would agree, I can just remove dynamic font change and let site owner to pick correct font.
But I would say that the font name suitable for Chrome should be somehow documented, isn't it?
May be in the the same editor.css

lib/ace/bidihandler.js
```diff
@@ -0,0 +1,338 @@
...
+ **/
+var BidiHandler = function(session) {
+    this.session = session;
+    this.bidiMap = {};
+    /* current screen row */
+    this.currentRow = null;
+    this.bidiUtil = bidiUtil;
+    /* Arabic/Hebrew character width differs from regular character width */
+    this.charWidths = [];
+    this.EOL;
```
ashensis: - bidiUtil references lightweight Bidi engine used to produce Logical <-> Visual maps. These maps are used to correctly calculate caret position.
- charWidths contains widths of various characters (English, Bidi, Zero width, Arabic Lam-Alef). Again, they are used to correctly calculate caret position.
- EOL - end of line character, it may vary in ACE. Line may be followed by end-of-line or end-of-file symbol (when show invisibles) and they are used in caret position calculations (sometimes EOL occurs inside the Bidi line, not at the end)

lib/ace/lib/bidiutil.js
```diff
@@ -0,0 +1,556 @@
...
+var RLO = 17; /* RLO						*/
+var BN = 18; /* Boundary Neutral			*/
+
+var TBBASE = 100, TB00 = 100, TB05 = 101, TB06 = 102, TB07 = 103, TB20 = 104, TBFB = 105, TBFE = 106, TBFF = 107;
+
+var MasterTable = [
+	/************************************************************************************************************************************/
+	/*		0		1		2		3		4		5		6		7		8		9		A	B		C		D		E		F	*/
+	/************************************************************************************************************************************/
+	/*0-*/	TB00,	L	,	L	,	L	,	L	,	TB05,	TB06,	TB07,	R	,	L	,	L	,	L	,	L	,	L	,	L	,	L	,
```
ashensis: As a matter of fact, approach implemented in codemirror is a rough one, it doesn't account for many outstanding cases.
However, after revising my classification tables I concluded that few of  them might be disposed off without much harm.
The most useful of existing tables may be replaced by regular expressions.
But I would prefer to retain partly, at least 2 tables, they might be compressed to incur less upload impact and shouldn't have performance impact as less probable to come in use often.
What do you think?

lib/ace/edit_session.js
nightwing: sorry, i meant line 255 https://github.com/ashensis/ace/blob/8b21722f4dff1e43a5c4debf066a54b0bcf6a9d3/lib/ace/edit_session.js#L255, session can be changed without calls to insert/remove. The only reliable way to handle that is to listen to the change event.

lib/ace/lib/bidiutil.js
nightwing: i think we should find a way to compress this significantly, otherwise we have to keep it as a separate extension, and not include in ace.js

lib/ace/edit_session.js
ashensis: Done.

lib/ace/layer/text.js
ashensis: Remove dynamic font change

lib/ace/lib/bidiutil.js
ashensis: I retained only 2 character classification tables (actually only part of each one)  and compressed them, the rest has been replaced by regular expressions

ashensis:
@nightwing I addressed all your comments, please have a look at submitted changes.

lib/ace/bidihandler.js
ebraminio: I guess the issue is `this.EOL;` is somehow NOOP here and it is not a function call like `this.EOL();`.

ebraminio: I guess the suggestion would be putting some default sane value here instead this NOOP.

ashensis: Did this for both  EOL and showInvisibles

adir01:
@nightwing Did @sashash answered all your comments ? Are you happy with his changes ?

nightwing:
looks good, i've merged with some trivial style changes, and will try using it for a while.
I amy have to turn this off by default, if performance difference is noticeable.

Thanks for all your hard work on implementing this, and your persistence and patience for getting this merged.

adir01:
Thank you @nigtwing

lyndondrake:
I am very eager to test this too. How do I access a build with this change included?

nightwing:
@lyndondrake clone this repository, install node, then run 
```
git clone https://github.com/ajaxorg/ace 
cd ace
npm install
node Makefile.dryice.js
```

or simply open https://rawgit.com/ajaxorg/ace/255c3a1793276b77eeba90e40df99c3ccc941780/kitchen-sink.html

shreevatsa:
This may not be exactly related to this PR but asking here as most knowledgeable people seem to be here:  

The spacing issues as described at https://github.com/ajaxorg/ace/issues/2667 and https://github.com/ajaxorg/ace/issues/3037 seem to still not be fixed even at https://rawgit.com/ajaxorg/ace/255c3a1793276b77eeba90e40df99c3ccc941780/kitchen-sink.html -- is that expected? For example if I put some random Indic text on a long line:
```
मम हि निष्कारणबान्धवं भवन्तमालोक्यैव दुःखान्धकारभाराक्रान्तेन महतः कालादुच्छ्वसितमिव चेतसा, श्रावयित्वा स्ववृत्तान्तमिमं सह्यतामिव गतः शोकः । दुःखितमपि जनं रमयन्ति सज्जनसमागमाः । परसुखोपपादनपराधीनश्च भवादृशां गुणोदयःऽ इत्युक्तवतीञ्चैनां चन्द्रापीडो ऽव्रवीत्-"भगवति! दर्शनात्प्रभृति परवानयं जनः कर्त्तव्येषु यथेष्टमशङ्किततया नियुज्यताम्ऽ इत्यभिधाय तया सहैवोदचलत् । क्रमेण च गत्वा हेमकूटमासाद्य गन्धर्वराजकुलम्, समतीत्य काञ्चनतोरणानि सप्तकक्षान्तराणि कन्यान्तःपुरद्वारमवाप । महाश्वेतादर्शनप्रधावितेन दूरादेव कृतप्रणामेन कनकवेत्रलताहस्तेन
```
and click somewhere in it and hit Enter, the line break gets inserted somewhere other than where the cursor was. Is there a way to resolve it by using a different font? Because it would be surprising if RTL text with varying char widths can work, but not left-to-right.

shreevatsa:
(Sigh. All I wanted was a textarea that can say safely support say 2 MB of plain text, with some minimal highlighting for `# section headings` and `**bold**` text. I expected that something like Ace might be overkill, but didn't expect this.)

So from what I can understand of this PR by looking at the code, it looks like there are two parts to it (and to supporting Arabic/Hebrew/Persian more generally):

* Accounting for the fact that not all characters (not codepoints: glyphs, graphemes, whatever) will have the same width, and
* Handling bidirectional issues (levels of right-to-left v/s left-to-right nesting, etc).

This PR does both, but what we'd need for supporting Indic scripts (or even characters like ⟦ ♞ ℝ as on the linked issues) is just the first part (correctly handling characters whose widths vary). 

Is there a way of extracting out just the first part, so that it would be more useful?

hartman:
So am I correct in understanding that glyph fallback replacement on Macs breaks this font width calculation, since it replaces with glyphs that are not monospaced ?

ashensis:
Hi @shreevatsa.
On addressing your comments, I should say that this PR was intended and tested primarily for Arabic & Hebrew languages that have, may be accidentally the common denominator in respect of character width.
I personally never dealt the Inidc.
Secondly, this PR doesn't take care of right-to-left direction (the issue I am planning to develop as a next step). So only left-to-right text orientation can be maintained for now.
The Bidirectional aspect of this PR lays with light weighted Bidi engine and it represents the bulk of code, while the maintaining variable character width constitutes relatively minor aspect. Besides, as far as I know, Ace team has plans to provide general support for variant width fonts.

ashensis:
Hi @hartman.
The "Monaco" monospaced font in fact isn't monospaced when it comes to Arabic and Hebrew glyph on Chrome engines.
Initially I developed automatic switch to Courier font for given line when typing of these characters occurs on Chrome. But this approach was out-ruled by reviewer due to this very issue of Mac having no Courier. 
However the site author is free to specify the default font in editor.css, so if one finds the monospaced font (in respect to Arabic/Hebrew) on Mac, he/she may fairly well to make corresponding site configuration.

shreevatsa:
Hi @ashensis -- from your comments it looks like you've done most of the work for supporting variable-width left-to-right scripts, and only some minor tweak remains. I am hopeful: if you have a few minutes could you look into the example I posted earlier (https://github.com/ajaxorg/ace/pull/3137#issuecomment-315412976 : you can install a Devanagari font like [Noto Sans Devanagari](https://www.google.com/get/noto/#sans-deva) if it doesn't render) and see why it has the issues I mentioned? My guess is that, if the text is simply detected / marked as being "bidi" (in terms of your code in this PR — even if it's strictly left-to-right), then everything would actually work, which would be great.

AFAIK the Ace team has no other plans in progress for variable-width characters, and as you've already done such a lot of work I feel there's probably only a minor change or two required for everything to work perfectly.

ashensis:
Hi @shreevatsa.

lastly I said that this PR was targeting only Arabic/Hebrew.
Please pay attention on the following (first) code line in bidihandler.js, 'updateCharacterWidths' method
this.updateCharacterWidths = function($fontMetrics) {
var _regularCharWidth = $fontMetrics.$measureCharWidth('a'), _bidiCharWidth =   $fontMetrics.$measureCharWidth('\u05d4');
It measures Hebrew letter Alef when calculating bidiCharWidth. This width is uniformly equal for all Arabic and Hebrew characters (except compound Arabic ones like LamAlef) on Monaco or Courier monospaced fonts.
I tried to measure widths of a few of Inidc characters posted by you, and they have variant widths, like:
$fontMetrics.$measureCharWidth('भ') == 11.4666666
$fontMetrics.$measureCharWidth('सा') == 15.666666
So I presume that in order to provide general support for none-monospaced fonts, more elaborate approach should be accepted, namely measuring the cursor offset from line beginning on every change caret position (which should take a toll on overall performance)

ashensis:
I didn't try to install Devanagari font but if you already have it installed, you can play around by measuring the widths of various Indic characters to see if it is truly monospaced from this respect.
If the answer would be positive, then the straightforward change may be done, say to provide a configuration paarmeter telling which script should be supported, Bidi, Indic or may be something else

ashensis:
And then to measure respecting character for script being configured.

shreevatsa:
@ashensis Thanks for that explanation, it was really helpful and cleared up my misunderstanding. I somehow seem to have missed that what's now supported is *monospaced* Arabic/Hebrew; I didn't even know those existed. :-) Thanks, guess the goal is further than I thought (as I'm not aware of monospaced Indic fonts, and what's really needed is variable-width support which as you say implies performance costs). Sorry for the noise arising from my misunderstanding.

Issue #3135, opened 2016-11-07
Limited marker functionality - no easy solution for popovers


oatssss:
Hi, I'm trying to extend the functionality of the editor's markers. Specifically, I'm trying to use react to achieve bootstrap-like popups when hovering over text markers. It seems there's no easy way to do this presently, but with a minor change to the Marker render process, this can be done in a fairly straightforward way.

The line in question is this one:
https://github.com/ajaxorg/ace/blob/ca0953efebb881f049cbf29e8a328aa5a33ad410/lib/ace/layer/marker.js#L96
and a PR with the full 2-line change required to make this work:
https://github.com/ajaxorg/ace/pull/3143/files

Here's a codepen showing off the kinds of things that can be done once this fix is in place (the codepen uses a custom build of Ace with insertAdjacentHTML).
https://codepen.io/oatssss/pen/oYxJQV

In this case, the change must be made so that modifications to the marker layer from within a dynamic marker's update, aren't eventually erased by the layer's assignment to innerHTML.

So how is this beneficial to Ace as a whole?
I find it's much more useful to developers integrating ace in their projects to be able to modify the editor's marker DOM in a manner of their choosing (react included), than being forced to use the current method of passing raw strings of html to a string builder.

Additionally, I would love to hear any other suggestions on how to achieve marker popups without modifying Ace's internals.

AlonBe:
I'm encountering the same issue.
@oatssss's PR is well explained

Any news regarding this issue?

elk941:
@oatssss have you looked at [this](https://codepen.io/oatssss/pen/oYxJQV/) ?

akoreman:
Closed as duplicate of #2720.

PR #3143, opened 2016-11-15
Switch to insertAdjacentHTML in marker rendering process

oatssss:
Associated issue: https://github.com/ajaxorg/ace/issues/3135

This change retains modifications made to the marker layer's DOM.

Currently, Ace renders its markers by concatenating each individual marker's HTML (string builder approach) and doing a final call to assign to the marker layer's innerHTML. This means that if any marker makes changes to the layer from within the update (for example by using appendChild or ReactDOM.render), the final assignment to innerHTML completely overrides the DOM change.

This 2-line change frees developers from being forced into rendering [dynamic]Markers using the string builder technique and permits other rendering options, most notably React.

Here is a codepen showing off what can be done with this change (it uses a custom Ace build based on the PR):
https://codepen.io/oatssss/pen/oYxJQV

### DIFF
--- a/lib/ace/layer/marker.js
+++ b/lib/ace/layer/marker.js
@@ -62,7 +62,7 @@ var Marker = function(parentEl) {
 
         this.config = config;
 
-
+        this.element.innerHTML = '';
         var html = [];
         for (var key in this.markers) {
             var marker = this.markers[key];
@@ -93,7 +93,7 @@ var Marker = function(parentEl) {
                 this.drawSingleLineMarker(html, range, marker.clazz + " ace_start" + " ace_br15", config);
             }
         }
-        this.element.innerHTML = html.join("");
+        this.element.insertAdjacentHTML('afterbegin', html.join(""));
     };
 
     this.$getTop = function(row, layerConfig) {


its-poole:
I agree that other approaches would be great if this is considered unidiomatic; if not, it indeed opens up a trove of possibilities.

xeviknal:
Hi there,

I am in the scenario where I want to add popovers for specific words/lines in a yaml'ed ACEditor. I see that the only point that prevents me to achieve that is this pending PR. Am I wrong? Is there any other way to achieve it?
@oatssss did you find another way to reach your solution? If yes, could you find it?

Cheers

o-ohst (approved):

oatssss:
Sorry I missed the notifications for this thread. I didn't find another way to do it. I'm not actively working on the project which initially required this anymore so I don't plan on committing further changes as I'm not so familiar in JS now, but one thing I noticed was that adding many dynamic markers had a significant hit to performance. I do not remember if it was because of this change, or because I was rendering using React, or if dynamic markers did not have good performance to begin with.

I caution anyone who uses this solution investigate if it performs well enough for their use case.

PR #3223, opened 2017-02-23
fix bug of operator '&' and '|' in lua mode

zoowii:
fix luamode bug of '&' and '|'
if (1 & 2) or (1 | 2) then
return
end


zoowii:
I'm sorry. I just remember you use the oxyc/luaparse repo for luaparse, bug can you update the dependency? too old version luaparse.js

PR #3225, opened 2017-02-23
add support for language glua. glua is a static type language based on

zoowii:
Add glua language support. glua is a static type language based on Lua. Its syntax like TypeScript and Lua

```
type Person = {
    name: string
    , age: int,
    hi: (string) => string
}

let a1 = (a: number, b: number) => do 
    print 'abc'
end

let f1 = (a: int, b: int) => a + b

let a2: string = '123'

type T3 = string | bool | 'hello'

var M = Contract<Storage>()

let cc = type(123)

function M:init(a: int)
    print("hello")
    print(a + 100)
    emit hello('abc')
end

type Person<T> = {
    name: string,
    country: T
}

type T1 = Person<string>
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,6 +4,7 @@ lib/ace/mode/html/*
 lib/ace/mode/javascript/*
 lib/ace/mode/json/*
 lib/ace/mode/lua/*
+lib/ace/mode/glua/*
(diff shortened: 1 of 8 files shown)


zoowii:
Now all checks passed

PR #3226, opened 2017-02-24
add support for language glua. glua is a static typed language based on Lua. Its syntax looks like TypeScript and Lua.

zoowii:
Add glua language support. glua is a static type language based on Lua. Its syntax like TypeScript and Lua

```
type Person = {
    name: string
    , age: int,
    hi: (string) => string
}

let a1 = (a: number, b: number) => do 
    print 'abc'
end

let f1 = (a: int, b: int) => a + b

let a2: string = '123'

type T3 = string | bool | 'hello'

var M = Contract<Storage>()

let cc = type(123)

function M:init(a: int)
    print("hello")
    print(a + 100)
    emit hello('abc')
end

type Person<T> = {
    name: string,
    country: T
}

type T1 = Person<string>
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,6 +4,7 @@ lib/ace/mode/html/*
 lib/ace/mode/javascript/*
 lib/ace/mode/json/*
 lib/ace/mode/lua/*
+lib/ace/mode/glua/*
 lib/ace/mode/php/*
 lib/ace/mode/xml/*
 lib/ace/mode/xquery/*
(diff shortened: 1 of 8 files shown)


ZeProf2Code:
Please include all the constants in the gLua available here : https://github.com/FPtje/Sublime-GLua-Highlight and all the snippets ;)

zoowii:
the glua mode is writen for lang `gualang.org`, which is only super set of `Lua5.3 + static type system`. 
and now
syntax `!=` is not supported.
bugs of `and` and `or` fixed.
snippets of glua added.

ZeProf2Code:
oh there is 2 gLua so haha I was talking about GmodLua you know (https://wiki.facepunch.com/gmod/)

PR #3229, opened 2017-02-28
Create composer.json

tylerssn:
Create composer.json so that package can be submitted to Packagist.


tylerssn:
Addresses issue https://github.com/ajaxorg/ace/issues/3154

tylerssn:
@nightwing - Closing this to clean up my profile open-PR list. But here it is if ever needed.

hxtree:
There doesn't appear to be a composer.json file. Ideally, this would be listed as an install method.

PR #3611, opened 2018-03-14
save + restore state when using heredocs

kevinushey:
This should fix an issue where modes that embed `sh` highlight rules can transition into an unexpected state after closing the heredoc.

The existing code seems to already attempt to restore an old saved state; however, as far as I can tell the state is never actually 'saved' when the heredoc is entered. That seems unintentional to me but it's possible I'm missing something?


kevinushey:
Actually, it seems like this might not be sufficient -- I'm going to dig a bit more ...

PR #3673, opened 2018-05-24
Added two tmthemes gruvbox dark (hard) and light (hard)

stuples:
Added two gruvbox themes: dark (hard contrast) and light (hard contrast). These are imported sublime themes from: https://github.com/Briles/gruvbox

To create the themes I followed the instructions from the wiki.


andrewnester:
Thanks for your contribution!

PR #3696, opened 2018-06-28
separate TS module declarations

43081j:
Fixes ajaxorg/ace-builds#133.

---

This outputs the module declarations into a separate file and references it in the main declaration file.

I haven't yet had chance to try it in my project but will do today if i get chance. If someone has chance to give it a go before then, that would be handy. Don't merge until it has been please.

cc @nightwing (not sure if you already started work on this, sorry if you had)

### DIFF
--- a/Makefile.dryice.js
+++ b/Makefile.dryice.js
@@ -112,9 +112,9 @@ function ace() {
 }
 
 function buildTypes() {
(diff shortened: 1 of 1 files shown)


43081j:
just tried it on my project and seems to work fine.

maybe @fr0 or @RoberttheProfessional can take a look seeing as they were able to reproduce the issue in their stack.

PR #3779, opened 2018-09-22
hide replace toggle & replace dialog when the editor is in read only mode

gnijuohz:
*Issue #, if available:*

#3773 

*Description of changes:*

When the editor in read only mode, the replace toggle button no longer shows in the search box and the replace dialog will be hidden as well.


gnijuohz:
Ah... Sorry I force pushed and now travis ci is failing. I'll close this one and submit a new pull request. 
😞

PR #3864, opened 2018-12-25
nameOverrides AutoHotKey to "AutoHotkey / AutoIt"

uqix:


uqix:
Does ace support AutoIt?

PR #4140, opened 2019-12-27
Improve Code Lens API

ignovak:
* Add provideCodeLenses callback argument to support asynchronous providers
* Add multiple providers support
* Rename misspelling "code lense" to "code lens"

### DIFF
--- a/demo/code_lens.html
+++ b/demo/code_lens.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>ACE Autocompletion demo</title>
+  <title>ACE Code Lens demo</title>
   <style type="text/css" media="screen">
     body {
         overflow: hidden;
@@ -48,8 +48,8 @@ require(["ace/ace", "ace/ext/code_lens"], function(ace, codeLens) {
     });
     editor.setOption("enableCodeLens", true);
 
-    codeLens.setCodeLensProvider(editor, {
-        provideCodeLenses: function(session) {
+    codeLens.registerCodeLensProvider(editor, {
+        provideCodeLenses: function(session, callback) {
             var p = [{
                 start: {row: 0},
                 command: {
@@ -110,7 +110,7 @@ require(["ace/ace", "ace/ext/code_lens"], function(ace, codeLens) {
                 });
 
             }
-            return p;
+            callback(p);
         }
     });
 
--- a/demo/code_lens.html
+++ b/demo/code_lens.html
@@ -7,8 +7,8 @@
     body {
         overflow: hidden;
     }
-    
-    #editor { 
+
+    #editor {
         margin: 0;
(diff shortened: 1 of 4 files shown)


ignovak:
NB: In the original code, there was a line `var codeLensProvider = session.codeLensProvider || session.$mode.codeLensProvider;`. Not sure how `session.$mode` should be taken into account with the new implementation. Please correct me if it's still needed.

lib/ace/ext/code_lens.js
```diff
@@ -155,30 +153,44 @@ function attachToEditor(editor) {
...
-        var lastDelta = session.$undoManager && session.$undoManager.$lastDelta;
-        if (lastDelta && lastDelta.action == "remove" && lastDelta.lines.length > 1)
-            return;
-        var row = session.documentToScreenRow(cursor);
-        var lineHeight = editor.renderer.layerConfig.lineHeight;
-        var top = session.getScrollTop() + (row - oldRow) * lineHeight;
-        session.setScrollTop(top);
+        var providersToWaitNum = session.codeLensProviders.length;
+        var lenses = [];
+        session.codeLensProviders.forEach(function(provider) {
```
keyboardDrummer: I don't think we need support for multiple code lens providers on the Ace side, so it might be better not to have this feature if we're not using it

lib/ace/ext/code_lens.js
```diff
@@ -155,30 +153,44 @@ function attachToEditor(editor) {
...
+            provider.provideCodeLenses(session, function(currentLenses) {
+                currentLenses.forEach(function(lens) {
+                    lenses.push(lens);
+                });
+                providersToWaitNum--;
```
keyboardDrummer: @nightwing are you open to Promises in Ace? It would be nicer to use `Promise.all` here.

lib/ace/ext/code_lens.js
```diff
@@ -155,30 +153,44 @@ function attachToEditor(editor) {
...
+                if (providersToWaitNum == 0) {
```
keyboardDrummer: I'm a little surprised by the `==` instead of `===`, but I see it's allowed from the eslint rules: https://github.com/ajaxorg/ace/blob/master/.eslintrc#L32

Those rules are very different from the ones we use in Cloud9: https://code.amazon.com/packages/AWSCloud9Public/blobs/mainline/--/packages/eslint-config/index.js

Is there are reason for this difference?

demo/code_lens.html
```diff
@@ -2,13 +2,13 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>ACE Autocompletion demo</title>
+  <title>ACE Code Lens demo</title>
```
keyboardDrummer: Putting the rename in this PR makes it difficult to review the other changes.

PR #4309, opened 2020-06-05
FIX monochar emojis by displaying them smaller

JeSuisUnCaillou:
*Issue #, if available:*
Emojis made of only one character don't have the correct cursor offset

*Description of changes:*
Managed to display them smaller so they correctly represent their width, monospace character wise.

It's not the best solution, I would have prefered to correctly offset the cursor, but didn't manage to do it.

### DIFF
--- a/lib/ace/css/editor.css
+++ b/lib/ace/css/editor.css
@@ -268,6 +268,11 @@ styles.join("\n")
     text-align: center;
 }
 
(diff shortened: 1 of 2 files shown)


JeSuisUnCaillou:
woops, sorry, didn't mean to merge into ace master. I close that.

Issue #4358, opened 2020-08-12
Markdown Syntax: Italics inconsistencies for intra word underlines


RickStrahl:
Looks like there's a slight issue with italic text in Markdown mode if there are underscores inside of words.

Try the following in Ace:

```markdown
_Syntax_highlighting_and_rendering_not_consistent_for_intra_words_underscores_

Syntax_highlighting_and_rendering_not_consistent_for_intra_words_underscores_
```

![image](https://user-images.githubusercontent.com/1374013/90075832-0b3f6600-dc9a-11ea-9203-0ec4189db9cc.png)

The underscores are being applied outside of word boundaries, and should only be applied to word boundaries.

A workaround is to escape the underscore based string in a inline code block:

```markdown
This is a constant `Syntax_highlighting_Error` and it renders properly.
```

which is what I would do, but several users of our tools actually flagged this issue at various times so apparently there is a use case for the funky syntax used. :smile: Linking the original reference in Markdown Monster that triggered this.

RickStrahl:
There's an update to this functionality in the latest update.

However, this may not be permanent. The above fix has issues on Internet Explorer/Web Browser control so it only partially works. The current build has this working as a custom fix to ACE, but if this doesn't get merged into the ACE master then this fix won't be available in the future.

PR #4361, opened 2020-08-15
Update italic highlighting rules to avoid intra word highlights in Markdown Mode. #4358

RickStrahl:
*Issue #4358

This address italic highlighting in Markdown Syntax.

*Description of changes:*

Changes the regex to check for word boundaries or beginning/end of line endings to avoid intra-word highlighting of italic text.

### DIFF
--- a/lib/ace/mode/markdown_highlight_rules.js
+++ b/lib/ace/mode/markdown_highlight_rules.js
@@ -148,7 +148,7 @@ var MarkdownHighlightRules = function() {
             regex : "([*]{2}|[_]{2}(?=\\S))(.*?\\S[*_]*)(\\1)"
         }, { // emphasis * _
             token : "string.emphasis",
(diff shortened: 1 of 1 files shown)


RickStrahl:
So, build is failing after PR. Not sure what this error means however. I was able to build locally without issues.

RickStrahl:
After some additional checking it looks like there's a problem with Internet Explorer failing to render this RegEx expression causing a markdown document to fail loading. Apparently the look behind regex is not supported in IE, but works in newer browsers.

Not sure what the solution to that issue is. Perhaps conditional code with a different string for IE check.

Unfortunately IE is what I currently need this to work in as this is a desktop application with an embedded WebBrowser control. 😏

RickStrahl:
So one of the Markdown tests is failing after this change. I've tried to figure out what that particular test checks for, but I can't see what the token / value combo accomplishes...

This change will change behavior in some cases, but legit italics should still work as before. Unfortunately I can't tell what the actual values are that are being tested in the failing test.

PR #4402, opened 2020-10-26
Added ES6 Default Parameters States

tmpmachine:
*Description of changes:*

Comparison between with default parameters states (left) & without (right) :

2020-10-26 17_23_44-TMPmachine_ Temporary Web App Code Editor

 It's important to write __default parameters__ states as its own group for two reasons :
- __function arguments__ states handles identifiers
- __default parameter__ states handles value

Below image is the proposed states for ES6 default parameters highlighting rules (JavaScript).

Untitled Diagram-Page-4

The actual states are more complex than this, and I know I'm missing something here. But the point is to show you the flow of the states.

### DIFF
--- a/lib/ace/mode/javascript_highlight_rules.js
+++ b/lib/ace/mode/javascript_highlight_rules.js
@@ -305,10 +305,43 @@ var JavaScriptHighlightRules = function(options) {
                 defaultToken: "string.regexp.charachterclass"
             }
         ],
+        "default_parameter": [
+            {
+                token : "string",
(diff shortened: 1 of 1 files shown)


andrewnester:
@tmpmachine thanks for contribution, unfortunately this PR breaks the tests, could you please take a look?

akoreman:
PR #4997 was merged which uses this PR as its basis. I'll go ahead and close this PR since it's effectively merged as part of the linked PR.

PR #4448, opened 2021-02-13
Fix vim's `dG` when performed over the whole range

### LINKED ISSUE #4411: Vim keybindings: `cG` does not delete file when run on the first line
`dG` works correctly: delete the whole file, starting from the current line.
However, `cG` is buggy and actually doesn't delete anything at all, it just inserts a new line above.

When run on any line other than the very first, it works fine.

_Originally reported by @jyn514 at https://github.com/integer32llc/rust-playground/issues/649_

bobrippling:
This fixes #4411, discovered on the rust playground (integer32llc/rust-playground#649).

We need to handle the special case of deleting the entire range of text, where we can't effectively move `anchor` to the previous line.

### DIFF
--- a/lib/ace/keyboard/vim.js
+++ b/lib/ace/keyboard/vim.js
@@ -2921,7 +2921,8 @@ dom.importCssString(".normal-mode .ace_cursor{\
           }
           var prevLineEnd = new Pos(anchor.line - 1, Number.MAX_VALUE);
           var wasLastLine = cm.firstLine() == cm.lastLine();
-          if (head.line > cm.lastLine() && args.linewise && !wasLastLine) {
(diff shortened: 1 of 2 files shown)


bobrippling:
@nightwing I'd like to contribute this but the contributor agreement seems to have disappeared. Is there anything that can be done about it?

image

nightwing:
Sorry, looks like the readme is a bit out of date, just adding the last line from https://github.com/ajaxorg/ace/blob/master/.github/PULL_REQUEST_TEMPLATE.md to the pull request description will be enough.

bobrippling:
Ah great - done, cheers

nightwing:
Thank you!

PR #4554, opened 2021-08-27
[WIP] replace importCssString with importCssStylesheet

dimovpetar:
*Issue:* #3260

*Description of changes:*
Instead of creating `style` element and embedding it in the `head`, now CSS files are preserved during build and then are imported as external styles.

**Important**
Current change affects only themes **ambiance** and **textmate**. If the approach is approved, I can continue with the rest of the themes.

### DIFF
--- a/Makefile.dryice.js
+++ b/Makefile.dryice.js
@@ -376,6 +376,18 @@ function buildAce(options, callback) {
     var snippetFiles = jsFileList("lib/ace/snippets");
     var modeNames = modeList();
 
+    copy("lib/ace/", getTargetDir(options) + "/", {
+        exclude: function (name) {
+            // skip folders
+            if (/^[^.]*$/.test(name)) {
+                return false;
+            }
+
+            return /^(?!.*css$).*$/.test(name)
+        },
+        include: /\.css$/
+    });
+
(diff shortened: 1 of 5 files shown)


dimovpetar:
Hello @nightwing ,

Thanks for your effort! We can load the CSS ourselves. 0d293163fcc664e51f7618abed7ac06f48d74781 already works well for us, would be nice to see it merged to the master.

Here are some minor improvents I would suggest:

- vim.js uses the `style` attribute, it can be replaced with class or `style.color` property https://github.com/ajaxorg/ace/blob/94422a4a892495564c56089af85019a8f8f24673/lib/ace/keyboard/vim.js#L4984
- static_highlight.js also sets the `style` attribute as string, not sure when this is used https://github.com/ajaxorg/ace/blob/94422a4a892495564c56089af85019a8f8f24673/lib/ace/ext/static_highlight.js#L63
- `ace.config.useStrictCSP(true);` can become `ace.config.set("useStrictCSP", true);` for consistency
- The generated CSS files can be minified

Do you plan to make a release with these CSP fixes?

dimovpetar:
Abandoning this in favor of #4558

PR #4618, opened 2022-01-19
replace escaped newline with actual newline when in regex mode

Eric162:
*Issue #, if available:* #2869

*Description of changes:* Right now, replacing a newline doesn't work in the ace editor. This allows replacing newlines when in regex replace mode.


Eric162:
This may not be the best place for the fix. Maybe in the search code would be better, but the general fix does work.

akoreman:
Closing as duplicate of #5027.

PR #4625, opened 2022-01-31
add highlighting support for Single-Assignment C

hv15:
*Description of changes:*

With this PR I'd like to include support to syntax highlight Single-Assignment C code, which is a functional array programming language (see https://sac-home.org for more info) with a very similar syntax to C/C++.

The mode rules are based off of the ace C/C++ mode file, but SaC is a subset of C so I've omitted a lot from it (it didn't make sense to extend it).


hv15:
has anyone had a chance to look at this yet?

PR #4651, opened 2022-03-10
Add support for min(), max() and var() to css linter

bogger33:
*Issue #, if available:* N/A

*Description of changes:*
This makes 2 small changes to the css linter, which help it support two commonly used modern css constructs:

1. add support for min()/max() for css properties accepting a length. while I was modifying/testing the code I also noticed that currently it accepted stuff like "calcasdhkjas()" as valid because it starts with "calc", which is accepted, so I fixed that as well.
2.  allow var() as a value anywhere, to fully support css custom properties. definition of them was already supported, but using them still gave errors.

### DIFF
--- a/lib/ace/mode/css/csslint.js
+++ b/lib/ace/mode/css/csslint.js
@@ -5914,7 +5914,7 @@ var Validation = module.exports = {
 
             // All properties accept some CSS-wide values.
             // https://drafts.csswg.org/css-values-3/#common-keywords
-            if (ValidationTypes.isAny(expression, "inherit | initial | unset")) {
+            if (ValidationTypes.isAny(expression, "inherit | initial | unset | var()")) {
                 if (expression.hasNext()) {
                     part = expression.next();
                     throw new ValidationError("Expected end of value but found '" + part + "'.", part.line, part.col);
--- a/lib/ace/mode/css/csslint.js
+++ b/lib/ace/mode/css/csslint.js
@@ -6238,7 +6238,7 @@ copy(ValidationTypes, {
         },
 
         "<length>": function(part) {
-            if (part.type === "function" && /^(?:-(?:ms|moz|o|webkit)-)?calc/i.test(part)) {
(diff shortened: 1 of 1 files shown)


bogger33:
Unfortunately what I previously did only worked on css style rules with a single value, upon further testing. It looks like it's not at all easy to add full support for css variables due to how the matcher works. What I've done now is add a hack that will simply stop error-checking a style rule once it encounters a css variable inside it.

This is a hack, in service of not throwing errors on encountering var(), but from what I can tell a proper solution would be quite involved. In my opinion, this is still better than the current state of things, since it will no longer throw false-positive errors, and there's a number of style rules that aren't validated as is anyway, because parsing them would be too complex (like e.g. border-image, background, etc. Anything with a 1 as the value in the Properties dictionary)

andrewnester:
Unfortunately, I think we need to support `var` properly and spend a bit more time for a better solution

bogger33:
Yeah, that doesn't surprise me, it is a hack after all (and even though other stuff like clip-path has a similar approach of just assuming things are good, at least in those cases it doesn't vary depending on what you input). 

Unfortunately supporting var properly will be quite complicated, it looks like. The problem is that there's some CSS rules that accept their parameters in an arbitrary order. E.g. for border, the valid input looks like this: ``"<border-width> || <border-style> || <color>"``

So currently, for something like ``border: 1px black solid;``, the parser will take the first input, ``"1px"``, cross the border width of the list, take the second input, cross the color of the list, and demand a border-style for the third input. If the list of inputs is empty after all that, the rule is valid.

Now if instead the rule looks like this: ``border: var(--border-width) black solid;``, at the point of the first input, we don't know the contents of the var (and since those contents are contextual since variables can be defined for any element and overwrite existing ones, we can't just build a database of them) - so we don't know whether the var contains a width, style or color. If we decide to just remove nothing from the list, we end up with a non-empty list, thus also not satisfying the criteria. And rules can get more complicated, like e.g. ``"auto | [ under || [ left | right ] ]"`` for text-underline-position.

So that was the point where I gave up and just added a hack to end the error spam.

PR #4705, opened 2022-05-19
Php ampersand fix

jeroenpost86:
*Issue #, if available:*

*Description of changes:*
Fixed the regex for T_AMPERSAND_FOLLOWED_BY_VAR_OR_VARARG and T_AMPERSAND_NOT_FOLLOWED_BY_VAR_OR_VARARG as this caused issues with the syntax checker


andrewnester:
@jeroenpost86 could you provide an example of code that caused errors before?

jeroenpost86:
> 

This some example code with the issues:

`<?php
    define("SPLIT", "\n===================\n\r");
    
    # for specific use cases. 
    
    function array_to_stdClass(&$array){
        $tmp = $array;
        $array = new stdClass();
        # false positive, works just fine
        foreach($tmp as $k => $v) $array->$k = (!is_object($v) && !is_array($v)) ? $v:array_to_stdClass($v);
        return $array;
    }
    function stdClass_to_array(&$stdClass){
        $tmp = $stdClass;
        $stdClass = array();
        # false positive, works just fine
        foreach($tmp as $k => $v) $stdClass[$k] = (!is_object($v) && !is_array($v)) ? $v:stdClass_to_array($v);
        return $stdClass;
    }
    `

It is because T_AMPERSAND_FOLLOWED_BY_VAR_OR_VARARG wasn't defined properly. 

Here is the same code running in an editor after the fix: https://onlinephp.io/c/2d00a

andrewnester:
I'm not sure I fully understand the issue. If I use this sample in kitchen-sink here https://ace.c9.io/build/kitchen-sink.html
everything parsed correctly and no errors reported. Then what this PR is trying to solve?

jeroenpost86:
> I'm not sure I fully understand the issue. If I use this sample in kitchen-sink here https://ace.c9.io/build/kitchen-sink.html everything parsed correctly and no errors reported. Then what this PR is trying to solve?

The kitchen sink is running an older version of the PHP parser. When I open the kitchen sink straight from the repo with `node ./static.js` the error shows up, but not on the link you posted.
Also, the kitchen sink doesnt support PHP8 yet, while the my last PR with PHP8 support was merged last week (what caused this specific issue).


PHP 8 stuff:
````
<?php

// Ace Editor PHP 8.1 demo

enum UserStatus: string
{
    case Pending = 'P';
    case Active = 'A';
    case Suspended = 'S';
    case CanceledByUser = 'C';

    public function label(): string
    {
        return match($this) {
            static::Pending => 'Pending',
            static::Active => 'Active',
            static::Suspended => 'Suspended',
            static::CanceledByUser => 'Canceled by user',
        };
    }
}

foreach (UserStatus::cases() as $case) {
    printf('<option value="%s">%s</option>'."\n", $case->value, $case->label());
}


class Service
{
    private Logger $logger;
    public readonly Status $status;
    final public const XX = "foo";

    public function __construct(
        Logger $logger = new NullLogger(), Status $status = new Status(),
    ) {
        $this->logger = $logger;
        $this->status = $status;
        $foo = $this->foo(...);

        $var ??= 0; // Syntax error, unexpected '='
        return $var + 1;
    }
    
    #[\Assert\All(
        new \Assert\NotNull,
        new \Assert\Length(min: 5))
    ]
    public string $name = '';

}

````

Screenshot of locally running kitchen sink:

````
git clone https://github.com/ajaxorg/ace.git ace
cd ace
node ./static.js
````
 

image

andrewnester:
Okay, got it, so it's essentially a followup for 8.1 update which was merged but not yet release. Thanks!

andrewnester (approved):

jeroenpost86:
> Okay, got it, so it's essentially a followup for 8.1 update which was merged but not yet release. Thanks!

Correct. Sorry for the confusion. I wrongly assumed the 8.1 update was already released :)

PR #4707, opened 2022-05-22
fix: Dracula theme constants should be pink, not purple

zardam:
In dracula theme, constants should be pink, not purple. See one example in IDLE :

image


andrewnester:
Based on Dracula theme specification, constant are supposed to be purple
https://spec.draculatheme.com/#sec-Constants

Closing, as current theme is according to spec

zardam:
Ok, my bad. At least two other editors are wrong then...

PR #4708, opened 2022-05-22
fix: Some Python syntax highlight enhancements

zardam:
fix: Some Python syntax highlight enhancements:
 * brackets should be matched independently
 * functions should be matched as functions

For example, with dracula theme, before:

image

After:

image

### DIFF
--- a/lib/ace/mode/python_highlight_rules.js
+++ b/lib/ace/mode/python_highlight_rules.js
@@ -170,6 +170,9 @@ var PythonHighlightRules = function() {
             token: "paren.rparen",
             regex: "[\\]\\)\\}]"
         }, {
+            token: ["keyword", "text", "entity.name.function"],
+            regex: "(def|class)(\\s+)(\\w+)"
+         }, {
             token: "text",
             regex: "\\s+"
         }, {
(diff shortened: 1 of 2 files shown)


lib/ace/mode/python_highlight_rules.js
```diff
@@ -165,10 +165,19 @@ var PythonHighlightRules = function() {
...
         }, {
             token: "paren.lparen",
-            regex: "[\\[\\(\\{]"
+            regex: "[\\(\\{]"
         }, {
             token: "paren.rparen",
-            regex: "[\\]\\)\\}]"
+            regex: "[\\)\\}]"
+        }, {
+            token: "paren.lbracket",
```
andrewnester, line 173: Could you explain why yo want to a new token here? `paren.lbracket` seems to be some new token you introduced and it won't be styled properly anyway.

zardam: You are right, I have removed these tokens from the pull request.

andrewnester (changes requested):
Please fix the test here, it should be `entity.name.function` after your change

https://github.com/ajaxorg/ace/blob/master/lib/ace/mode/_test/tokens_python.json#L336

zardam:
I have updated the pull request.

andrewnester (approved):

andrewnester:
Merged, thanks for the contribution!

PR #4726, opened 2022-05-30
Adds missed functions into `php_completions.js`

WinterSilence:
*Issue #, if available:*
#4546

*Description of changes:*
Adds missed functions (only popular)

- [x] By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


lib/ace/mode/php_completions.js
```diff
@@ -9271,7 +9271,262 @@ var functionMap = {
     "zlib_get_coding_type": [
         "string zlib_get_coding_type()",
         "Returns the coding type used for output compression"
-    ]
+    ],
+    // The new functions
+    // TODO: grab stubs and summary descriptions from php.net
```
andrewnester, line 9276: Can we add descriptions and arguments in this PR as well? I understand it's quite some work but we need to do it right :)

andrewnester (changes requested):

lib/ace/mode/php_completions.js
WinterSilence: @andrewnester no problem, use `getinfo.php` get info about function, script return array(json) in autocompletion format.

`getinfo.php`:
~~~php
// HTTP: 'getinfo.php?fn=function_name'
// CLI: 'php --file getinfo.php function_name'
$func = isset($_GET, $_GET['fn']) ? $_GET['fn'] : $_SERVER['argv'][1];
$html = file_get_contents('https://www.php.net/manual/en/function.' . str_replace('_', '-', $func));
$response = [];
preg_match('/\<div class="methodsynopsis dc-description"\>.+?\<\/div\>/uis', $html, $matches);
$response[] = trim(strip_tags($matches[0]));
preg_match('/\<span class="dc-title"\>.+?\<\/span\>/ui', $html, $matches);
$response[] = trim(strip_tags($matches[0]));
echo json_encode($response, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
~~~

andrewnester: But can we add this to this PR so we merge new functions with proper descriptions?

WinterSilence: @andrewnester yes, we can, but my part of teamwork is over

andrewnester: okay, thanks, I'll leave this in pending for now and see if we can take it over from here later

Issue #4687, opened 2022-05-05
Highlighter highlights all lines inside folded regions 


annikovk:
I have serveral big folded regions (several thousand lines) stay one after another: 
![image](https://user-images.githubusercontent.com/12910542/166903198-43021e16-8793-45a7-ac62-27fe8c968657.png)

Once I search something, highlighter is invoked to highlight all of the found strings: https://github.com/ajaxorg/ace/blob/master/lib/ace/ext/searchbox.js#L266

During the highlighting process, it takes editor's visible rows, and draws single line markers for every found range within visible rows. 

So, the problem is the fact that Ace thinks that rows inside folded regions are visible and tries to add markers for all of them:

![image](https://user-images.githubusercontent.com/12910542/166905273-63d5983f-6c6e-4a09-9b34-997d06cf4a73.png)

This causes UI freezes during the search: https://github.com/annikovk/IntelliJ-Log-Analyzer/issues/11
The only workaround I found is to unfold all regions, which is not acceptable.

andrewnester:
@annikovk could you please provide a minimal code so we can easier reproduce this issue? Thank you

annikovk:
@andrewnester have you been able to reproduce that? 
Is it a confirmed bug now?

andrewnester:
@annikovk you mentioned
> The only workaround I found is to unfold all regions, which is not acceptable.

How does unfolding help in this case? These selected word elements are still there, right?

annikovk:
I am trying to eliminate  the freeze. And the freeze happens in case there are thousands of selected words inside the folded region, which is inside visible area of the editor

unfolding helps to move selected elements out of visible region, so they are removed from DOM

Let me know if it is needed to describe that in more details with screenshots.

andrewnester:
Okay, got it!
I guess we indeed should just have 1 selected word dom element when the code is collapsed, there is no need to have them overlap.
I guess the reason we have is not to add / remove dom elements when code folds for performance reasons but it also seems like it already causes performance issues :)

I’ll try to work on PR for this

andrewnester:
@annikovk could you please check if your issue is solved here? https://raw.githack.com/andrewnester/ace/bug/4687/kitchen-sink.html

annikovk:
Yes, 100% working as expected. Thank you!

andrewnester:
We just merged the PR and released new 1.5.2 version containing the fix

PR #4727, opened 2022-05-30
fix: Do not render selected word markers for the same range multiple times

andrewnester:
*Issue #, if available:*
https://github.com/ajaxorg/ace/issues/4687

*Description of changes:*
Avoid rendering multiple markers at the same range to avoid browser freeze when a lot of search values found in folded code.

### DIFF
--- a/lib/ace/search_highlight.js
+++ b/lib/ace/search_highlight.js
@@ -56,7 +56,7 @@ var SearchHighlight = function(regExp, clazz, type) {
         if (!this.regExp)
             return;
         var start = config.firstRow, end = config.lastRow;
-        var renderedMarkerRanges = new Set();
+        var renderedMarkerRanges = {};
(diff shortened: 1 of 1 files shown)


nightwing (commented):
both calling toScreenRange, and searching for the word in a large number of lines are slow. It may be better to either use a loop skipping folded lines  https://github.com/ajaxorg/ace/blob/master/lib/ace/layer/text.js#L292 or use lines data from text layer.

In the long run it may be worth to unify all these loops into something similar to getVisibleRanges of monaco.

andrewnester:
@nightwing I don't think this CR changes how often we call `toScreenRange` so from this perspective it should stay the same but I agree that generally we need to revisit this and just skip folded areas

nightwing (approved):

PR #4810, opened 2022-05-31
Add class to tooltip DOM element distinguish errors from warnings

BenSouchet:
### Description of changes:
Adding required lines to add a CSS class to the tooltip element to distinguish errors from warnings.

Test by building ace from my branch, it's working like a charm:
Screenshot 2022-05-31 at 20 45 56

Not sure about if it the proper way to reset the `className` with an hardcoded value tho'

### DIFF
--- a/lib/ace/mouse/default_gutter_handler.js
+++ b/lib/ace/mouse/default_gutter_handler.js
@@ -87,6 +87,7 @@ function GutterHandler(mouseHandler) {
         tooltipAnnotation = annotation.text.join("<br/>");
 
         tooltip.setHtml(tooltipAnnotation);
+        tooltip.setClassName(annotation.className.trim());
         tooltip.show();
         editor._signal("showGutterTooltip", tooltip);
         editor.on("mousewheel", hideTooltip);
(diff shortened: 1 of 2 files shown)


lib/ace/mouse/default_gutter_handler.js
```diff
@@ -87,6 +87,7 @@ function GutterHandler(mouseHandler) {
         tooltipAnnotation = annotation.text.join("<br/>");
 
         tooltip.setHtml(tooltipAnnotation);
+        tooltip.setClassName(annotation.className.trim());
```
andrewnester, line 90: It might happen that className is undefined or null when annotation type is custom for example, so we should account for this too

andrewnester (changes requested):
Left some comments, but please also add a test for this functionality

BenSouchet:
Last point it seems there isn't any test for the `tooltip.js` file (no `tooltip_test.js`) is there a way to get or generate a starting point for that test file? 

Maybe this merge request isn't the best place for the creation and writing of all the tests of the `tooltip.js` file.

(I think I can manage to add one test to an existing test file but create the whole file with the setup/init process It's would be too difficult for me 🙁 )

andrewnester:
@BenSouchet sure, no worries about the test, I really thought we had one for tooltip already

andrewnester (approved):

andrewnester:
Merged, thanks for contribution!

Issue #4809, opened 2022-05-31
Annotations: duplicate annotations in CSS mode


BenSouchet:
### Describe the bug

In CSS mode if we enter this:
```CSS
h2 {
    

```
The newlines are mandatory to ensure the bug to occurs.

We get 2x the exact same "error" annotation (same `text`, `row`, `col`, `rule`, ...).

I currently use the 1.5.0 `ace-min` version available [in the build repo](https://github.com/ajaxorg/ace-builds/tree/master/src-min).

### Expected Behavior

Only one annotation. Not twice the same one.

### Current Behavior

Currently we get twice the exact same annotation:
```
{row: 2, column: 0, text: "Unexpected token 'null' at line 3, col 1.", type: 'error', rule: 'Parsing Errors'}
{row: 2, column: 0, text: "Unexpected token 'null' at line 3, col 1.", type: 'error', rule: 'Parsing Errors'}
```

### Reproduction Steps

Initialize an Ace editor with CSS mode, then enter these lines:
```CSS
h2 {
    

```
The newlines are required.

Now if you console log the annotations of the current Editor session you will get:
```
{row: 2, column: 0, text: "Unexpected token 'null' at line 3, col 1.", type: 'error', rule: 'Parsing Errors'}
{row: 2, column: 0, text: "Unexpected token 'null' at line 3, col 1.", type: 'error', rule: 'Parsing Errors'}
```
Twice the same with the exact same line and column.

### Ace Version

1.5.0

BenSouchet:
Additional info: I have tested on version `1.5.3` just in case, and the issue is the same.

BenSouchet:
@andrewnester I tried to investigate a bit 🙂
After a bit of time and a lot of breakpoints I understood that two annotations are registered because when verifying the CSS code in the `worker-css.js`:
- There is a first `SyntaxError` catch in `_readDeclarations()` (called by `_ruleset()`, itself called by `_stylesheet()`).
- Still in `_readDeclarations()` there is these lines:
```javascript
else if (tt !== Tokens.RBRACE) {
    throw ex;
}
```
- So the exception is forwarded and catch again by the caller `_stylesheet()`, this will generate the second msg.

When the exception is caught a `reporter.error(` is called to add an error, this is why there is two times the exact same annotation.

## The questions are:
- Why in `_readDeclarations()` the exception is consumed and throw to be consumed again ?
- Why the token value is `null`? Since the true error is `missing RBRACE`.

[...]

PR #4816, opened 2022-06-02
Fix css EOF duplicate errors

BenSouchet:
Issue #4809 | Annotations: duplicate annotations in CSS mode

This Pull request fix a bug with the CSS Parser (file `csslint.js`), the bug is simple :
If a declaration is open `{` and after that the parser encounter the end of the file `EOF` it will throw twice a wrong error (`Unexpected token 'null' at line 3, col 1.`).

I imagine this edge case has never been considered, and therefore taken into account in the code.

### Before / After
Before:
Before

After:
After

### DIFF
--- a/lib/ace/mode/css/csslint.js
+++ b/lib/ace/mode/css/csslint.js
@@ -1879,9 +1879,8 @@ Parser.prototype = function() {
                     this._readWhitespace();
                 } else {
                     var tt = tokenStream.peek();
-                    if (tt === Tokens.EOF || tt === Tokens.RBRACE) {
-                        // Encountered when there are no more properties.
-                    } else {
+                    // If this isn't an right brace or the end of the file throw an SyntaxError.
+                    if (tt !== Tokens.EOF && tt !== Tokens.RBRACE) {
                         this._unexpectedToken(tokenStream.LT(1));
                     }
                 }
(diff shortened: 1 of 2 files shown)


lib/ace/mode/css/csslint.js
```diff
@@ -1877,10 +1877,13 @@ Parser.prototype = function() {
 
                     value = new PropertyName(propertyName, hack, line || token.startLine, col || token.startCol);
                     this._readWhitespace();
-                } else if (tokenStream.peek() === Tokens.RBRACE) {
-                    // Encountered when there are no more properties.
                 } else {
-                    this._unexpectedToken(tokenStream.LT(1));
+                    var tt = tokenStream.peek();
+                    if (tt === Tokens.EOF || tt === Tokens.RBRACE) {
```
andrewnester, line 1882: it's a bit strange to have an empty if block, better inverse the condition to 
```
if (tt !== Tokens.EOF && tt !== Tokens.RBRACE)
```

andrewnester (changes requested):

lib/ace/mode/css/csslint.js
BenSouchet: Yes I see that but I didn't wanted to change this odd syntax to preserve the comment logic. I just push the modification.

BenSouchet:
By the way there is lot of odd elements in the `csslint.js` file, like these empty `else` statement:
https://github.com/ajaxorg/ace/blob/3f934510514a25c53edf64bb80911a96b7133908/lib/ace/mode/css/csslint.js#L4902
https://github.com/ajaxorg/ace/blob/3f934510514a25c53edf64bb80911a96b7133908/lib/ace/mode/css/csslint.js#L4910

If I had motivation and time I would try to fix a bit the file, but currently I will focus on fixing bugs I encounter.

Issue #4204, opened 2020-03-04
RTL Editing and Extra Characters in saved Text


RickStrahl:
I'm finally getting around to supporting RTL in one of my tools that uses ACE, and I'm happy to see that RTL is mostly working now in recent versions. Awesome.

Everything looks good in the editor, however when I take the content of an edited document that has a part of the document set to RTL, the content includes extra characters presumably to mark the text as RTL. It looks like any text following a line break ends up with an extra `20 2B` characters.

I'm using ACE in a Markdown editor and when editing the text as RTL these extra chars are breaking the Markdown down text which are treated like white space - the end result is that any paragraph breaks are not recognized and treated like line continuations.

Here's what this looks like in the editor and with the preview for the captured Markdown text from ACE:

![image](https://user-images.githubusercontent.com/1374013/75835678-a1727600-5d63-11ea-8d86-bc0f1bebd989.png)

Notice that both top and bottom sections are basically the same text, but the lower renders as a single paragraph while the top correctly renders the breaks.

It appears that ACE is embedding extra characters into the document to delineate the RTL text. This seems to work in that I can close the document and reopen it and ACE will automatically treat the RTL text with the correct alignment and text direction in the editor.

Looking at the Hex content I can see the extra characters:

![image](https://user-images.githubusercontent.com/1374013/75835779-e4cce480-5d63-11ea-9e41-fa5964bdd2c8.png)

in the RTL content.

Question: Is this the expected behavior and if so is this standard or some convention in ACE?

I apologize for the noobie questions around RTL - I have a hard time making sense of how this actually should work for an edited document given that other editors just display LTR no matter what. 

Any input relating to RTL with ACE would be welcome... if you're using it how do you store and manage the content and does this different content formatting affect you in any way. The only work around I see at this time is to potentially strip out the extra characters for rendering, but there should be less brute force way?

PR #4835, opened 2022-07-07
Fix undefined error thrown when using soft wraps and RTL font

dntrkv:
*Issue #:*
N/A

*Description of changes:*
Our organization ran into an issue where we see an undefined error being thrown when users attempt to use RTL fonts in an editor with soft wraps enabled.

I found the offending line of code in the bidihandler.js file where there is a conditional that checks for truthiness of the `splits` variable but then directly below/outside that block, the code attempts to access `splits.length` without first checking for truthiness.

Side note: I've found the RTL functionality in Ace to be rather buggy with some weird indentations that appear when writing multiple lines using RTL mode + RTL font. Is there an issue that outlines the current state of support for RTL languages?

### DIFF
--- a/lib/ace/bidihandler.js
+++ b/lib/ace/bidihandler.js
@@ -160,9 +160,11 @@ var BidiHandler = function(session) {
                 } else {
                     this.line = this.line.substring(0, splits[splitIndex]);
                 }
+
+                if (splitIndex == splits.length) {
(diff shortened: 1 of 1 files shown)


andrewnester:
I don't think we have an active tracking issue for RTL support, we had something previously but it was closed due to inactivity
https://github.com/ajaxorg/ace/issues/4204

Feel free to great one which describes the issues you see

PR #4905, opened 2022-08-24
Custom scrollbar with gutter indicators

mkslanc:
*Issue #, if available:*

*Description of changes:*
Adds custom scrollbar with indicators for annotations

### DIFF
--- a/src/layer/decorators.js
+++ b/src/layer/decorators.js
@@ -1,0 +1,149 @@
+"use strict";
+var dom = require("../lib/dom");
+var oop = require("../lib/oop");
+var EventEmitter = require("../lib/event_emitter").EventEmitter;
+
+var Decorator = function (parent, renderer) {
+    this.canvas = dom.createElement("canvas");
+    this.renderer = renderer;
+    this.pixelRatio = 1;
+    this.maxHeight = renderer.layerConfig.maxHeight;
+    this.lineHeight = renderer.layerConfig.lineHeight;
+    this.canvasHeight = parent.parent.scrollHeight;
+    this.heightRatio = this.canvasHeight / this.maxHeight;
+    this.canvasWidth = parent.width;
+    this.minDecorationHeight = (2 * this.pixelRatio) | 0;
+    this.halfMinDecorationHeight = (this.minDecorationHeight / 2) | 0;
+
+    this.canvas.width = this.canvasWidth;
+    this.canvas.height = this.canvasHeight;
+    this.canvas.style.top = 0  + "px";
+    this.canvas.style.right = 0 + "px";
+    this.canvas.style.zIndex = 7 + "px";
+    this.canvas.style.position = "absolute";
+    this.colors = {};
+    this.colors.dark = {
+        "error": "rgba(255, 18, 18, 1)",
+        "warning": "rgba(18, 136, 18, 1)",
+        "info": "rgba(18, 18, 136, 1)"
+    };
+
+    this.colors.light = {
+        "error": "rgba(255, 18, 18, 0.7)",
+        "warning": "rgba(18, 136, 18, 0.7)",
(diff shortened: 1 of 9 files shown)


src/layer/decorators.js
```diff
@@ -0,0 +1,143 @@
...
+                        yCenter = this.halfMinDecorationHeight;
+                    }
+                    else if (yCenter + this.halfMinDecorationHeight > this.canvasHeight) {
+                        yCenter = this.canvasHeight - this.halfMinDecorationHeight;
+                    }
+                    y1 = Math.round(yCenter - this.halfMinDecorationHeight);
+                    y2 = Math.round(yCenter + this.halfMinDecorationHeight);
+                }
+
+                switch (annotations[i].type) {
```
andrewnester, line 103: can we instead simplify this to

```
ctx.fillStyle = colors[annotations[i].type] || null

```

src/layer/decorators.js
```diff
@@ -0,0 +1,143 @@
...
+        function compare(a, b) {
+            if (a.typeNumber < b.typeNumber) return -1;
+            if (a.typeNumber > b.typeNumber) return 1;
+            return 0;
+        }
+
+        var annotations = this.renderer.session.$annotations;
+        ctx.clearRect(0, 0, this.canvas.width, this.canvas.height);
+        if (annotations) {
+            annotations.forEach(function (item) {
```
andrewnester, line 68: can we do something like

can we instead simplify this to

```
item.typeNumber = typeNumbers[item.type] || null

```

src/layer/decorators.js
```diff
@@ -0,0 +1,143 @@
...
+            if (allLineHeight < this.canvasHeight) {
+                this.heightRatio = 1;
+            } else {
+                this.heightRatio = this.canvasHeight / this.maxHeight;
+            }
+        }
+        var ctx = this.canvas.getContext("2d");
+
+        function compare(a, b) {
+            if (a.typeNumber < b.typeNumber) return -1;
```
andrewnester, line 60: I'd better rename it `typeNumber` -> `priority`

src/layer/decorators.js
```diff
@@ -0,0 +1,143 @@
...
+        if (cursor) {
+            let compensateFold = this.compensateFoldRows(cursor.row, foldData);
+            let currentY = Math.round((cursor.row - compensateFold) * this.lineHeight * this.heightRatio);
+            ctx.fillStyle = "rgba(0, 0, 0, 0.5)";
+            ctx.fillRect(0, currentY, this.canvasWidth, 2);
+        }
+
+    };
+
+    this.compensateFoldRows = function (currentRow, foldData) {
```
andrewnester, line 127: It seems like this logic is not covered with tests, can we add tests for this?

src/layer/decorators.js
```diff
@@ -0,0 +1,143 @@
...
+                        item.typeNumber = 3;
+                        break;
+                }
+            });
+            annotations = annotations.sort(compare);
+            var foldData = this.renderer.session.$foldData;
+
+            for (let i = 0; i < annotations.length; i++) {
+                let currentRow = annotations[i].row;
+                let compensateFold = this.compensateFoldRows(currentRow, foldData);
```
andrewnester, line 86: I'd rename currentRow to just row to avoid confusion with current cursor position

src/layer/decorators.js
andrewnester, line 127: I'd rename currentRow to just row to avoid confusion with current cursor position

andrewnester (changes requested):

andrewnester (approved):

src/layer/decorators.js
mkslanc: Yes, that's definitely better, corrected!

Issue #3262, opened 2017-04-06
Tooltip position issue inside a element with css "transform" property.


adolli:
Hi:

I use ace/tooltip in the following way.
```js
var Tooltip = require("ace/tooltip").Tooltip;
tooltip = new Tooltip(editor.container);
tooltip.setPosition(x, y);  // x, y relate to window top left
```
html likes the following
```html
<div class='.panel' style='transform: translate(-70px, -419px);'>
    <div id='ace-editor'></div>
</div>
```
**The tooltip dose not display related to the window, but the `.panel`.** I have checked out that w3c doc(I saw others comment this issue) says `posision:fixed` will lose efficacy when using css transform. Any ideas about how to make a tricky way to handle this situation?

[bug rel](https://bugs.chromium.org/p/chromium/issues/detail?id=20574)

Thanks.

adolli:
I finally found a solution. I just replace `tooltip = new Tooltip(editor.container);` with `tooltip = new Tooltip(document.getElementsByTagName('body')[0]);` to make `.tool_tip` element be the child of <body> avoiding being inside some <div>. The positions are now all correct.

Issue #2629, opened 2015-08-12
Position of tooltip is of on transformed elements


Daanoz:
The position of the tooltip is off, when the editor has a parent that uses the css transform property to position itself. For example modal dialogs use this to center the content.

A codepen to illustrate the problem: http://codepen.io/anon/pen/doLPOP

The transform property creates it's own coordinate system, causing 'position:fixed' to be relative to the container instead of the page.
A solution would be to add to tooltip to the body element of the page instead of contained within the editor element or to calculate a relative offset to the editor instead of using the events clientX / clientY.

Neill83:
can someone check this and fix it please? @nightwing ?

tsmirnov:
It seems, that it's pretty hard issue. Nobody have fixed it for 2 years. What if we just add config option, like "disableTooltip"? I think it is easy and quick solution. I can work on PR. What do you think?

mczepiel:
I notice  the autocomplete "tooltips" don't exhibit this behavior as they are presented in a DOM node outside of the editor's container.

In addition to not being subjected to any transformations on that container, this also means they appear in a different stacking context, which may or may not be desirable depending on your scenario.

Anyway, there's no easy way to influence the `MouseHandler` or override how it creates a `GutterHandler`, or how that creates a `GutterTooltip`.

Ideally, you could inject your own objects somewhere in that chain and re-root your tooltips.

As a quick hack, you _can_ alter the base `Tooltip.prototype.$init` and set `this.$parentNode` right before the tooltip creates its own node and appends it to said container.

e.g.
```javascript
const init = ToolTip.prototype.$init;
const tooltipContainer = document.body; // or something;

Tooltip.prototype.$init = function () {
  // if (constructor.name === 'GutterTooltip') or however you want to decide this
  this.$parentNode = tooltipContainer;
  init.call(this);
}
```

Not a great solution, but possibly a way to see if it's worth pursuing work to make it easier to influence the parentContainer that's hard-coded right now.

EmaGht:
Have the same issue here... is there something Work in Progress or a workaround to prevent this?
Thanks

blongstreth:
So, I ran into this very issue using Angular.js UI bootstap modal.  I tried various things and found the workaround proposed by @mczepiel works pretty well, so thank you!.  Also, I noticed that the AcePopup component is not placed in the editor container, but rather the body or document element (similar to this workaround).  

Example usage in ext-language_tools.js:
[...]

PR #4920, opened 2022-09-06
Append tooltip to body to ensure good positioning

AprilArcus:

The engineer who originally accomplished this work has long since departed my team, so forgive me for this somewhat rote transcription of his working notes.

> Tooltip positioning was being messed up in our modal because of the position on the modal's container.  Skip the issue by appending the tooltip to the body instead of the editor container so that the mouse coords it uses to compute position are always correct.

This issue appears to be related to [ajaxorg/ace#2629](https://github.com/ajaxorg/ace/issues/2629) and [ajaxorg/ace#3262](https://github.com/ajaxorg/ace/issues/3262).

### DIFF
--- a/src/mouse/default_gutter_handler.js
+++ b/src/mouse/default_gutter_handler.js
@@ -7,7 +7,7 @@ var Tooltip = require("../tooltip").Tooltip;
 function GutterHandler(mouseHandler) {
     var editor = mouseHandler.editor;
     var gutter = editor.renderer.$gutterLayer;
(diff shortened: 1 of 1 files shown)


andrewnester:
I don't think this change work well when we have multiple editors on the same page, what do you think @nightwing ?

akoreman:
Thanks for your contribution, unfortunately, in addition to the problem mentioned above, there are also some styling issues when tooltips are appended to the body.

PR #4947, opened 2022-10-10
Feat: Matching tags for highlighting/folding

mkslanc:
*Issue #, if available:* #4282 #2424

*Description of changes:*
- `getMatchingTags` method returns both open and close tag Ranges
- Unifies three different implementations of matching tag from `jumptomatching`, `xml folding` and `$highlightTags`
- `findMatchingTags` implementation for vim
- `jumpToMatching` better behaviour, when cursor is on closing tag
- Combines `$highlightTags` and `$highlightBrackets`
- Highlights both tag names
- Highlights indent guide, when cursor is on tag name 

### DIFF
--- a/src/edit_session/bracket_match.js
+++ b/src/edit_session/bracket_match.js
@@ -219,5 +219,190 @@ function BracketMatch() {
         
         return null;
     };
+
+    /**
+     * Returns [[Range]]'s for matching tags and tag names, if there are any
+     * @param {Position} pos
+     * @returns {{closeTag: Range, closeTagName: Range, openTag: Range, openTagName: Range} | undefined}
+     */
+    this.getMatchingTags = function (pos) {
+        var iterator = new TokenIterator(this, pos.row, pos.column);
+        var token = iterator.getCurrentToken();
+        var found = false;
+        var backward = false;
+        if (token && token.type.indexOf('tag-name') === -1) {
+            do {
+                if (backward) token = iterator.stepBackward(); else token = iterator.stepForward();
+                if (token) {
+                    if (token.value === "/>") backward = true;
+                    if (token.type.indexOf('tag-name') !== -1) {
+                        found = true;
+                    }
+                }
+            } while (token && !found);
+        }
+        if (!token) return;
+
+        var tag = token.value;
+        var currentTag = token.value;
+        var depth = 0;
+        var prevToken = iterator.stepBackward();
+
+        if (prevToken.value === '<') {
+            //find closing tag
+            var openTagStart = new Range(iterator.getCurrentTokenRow(), iterator.getCurrentTokenColumn(),
+                iterator.getCurrentTokenRow(), iterator.getCurrentTokenColumn() + 1
+            );
+            token = iterator.stepForward();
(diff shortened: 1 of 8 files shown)


src/edit_session/bracket_match.js
```diff
@@ -219,5 +219,190 @@ function BracketMatch() {
...
+     * @returns {{closeTag: Range, closeTagName: Range, openTag: Range, openTagName: Range} | undefined}
+     */
+    this.getMatchingTags = function (pos) {
+        var iterator = new TokenIterator(this, pos.row, pos.column);
+        var token = iterator.getCurrentToken();
+        var found = false;
+        var backward = false;
+        if (token && token.type.indexOf('tag-name') === -1) {
+            do {
+                if (backward) token = iterator.stepBackward(); else token = iterator.stepForward();
```
andrewnester, line 235: is there a possibility that we enter an endless loop here? maybe we should mark visited tokens so we break in this case?

src/edit_session/bracket_match.js
```diff
@@ -219,5 +219,190 @@ function BracketMatch() {
...
+            var openTagStart = new Range(iterator.getCurrentTokenRow(), iterator.getCurrentTokenColumn(),
+                iterator.getCurrentTokenRow(), iterator.getCurrentTokenColumn() + 1
+            );
+            token = iterator.stepForward();
+            var openTagName = new Range(iterator.getCurrentTokenRow(), iterator.getCurrentTokenColumn(),
+                iterator.getCurrentTokenRow(), iterator.getCurrentTokenColumn() + token.value.length
+            );
+            iterator.stepBackward();
+            var foundOpenTagEnd = false;
+            do {
```
andrewnester, line 262: this loop and function seems a bit too complex to understand. Could you refactor it out into separate methods / functions which are easier to read?

andrewnester (changes requested):
Functionality wise everything looks good, but I left some comments regarding refactoring the code we touched

src/edit_session/bracket_match.js
mkslanc: Actually, I don't see possibility of endless loop here, because of loop end condition - checking availability of `token`.   This `backward` check set only for one case, when we are in self-closing tag in between tag name and tag closing. I could improve this by remembering initial token and changing token to it in case we changed direction (direction is changing only once) or just put some comment, explaining why we need this check.

src/edit_session/bracket_match.js
mkslanc: I used logic from `$highlightTags` method and evolved it to support both closing and opening ranges. Will try to break it into some logical blocks. Thank you!

src/edit_session/bracket_match.js
andrewnester: @mkslanc I guess having a comment can work as well so it's easier to understand the context here

src/edit_session/bracket_match.js
mkslanc: I hope that would work

andrewnester (approved):

andrewnester:
Thank you! Approved and merged

PR #4949, opened 2022-10-11
Add ability to configure exact match in autocomplete.

tcarro927:
Need a way to specify that autocomplete uses exact matching when suggesting options.

I have a use case where the user types 'at' and autocomplete displays the following options:

atsec
attempt
active
datacenter

Users only want to see options that begin with the letters 'at' not options that contain a 'a' somewhere followed by a 't'. Setting the exactMatch property of the Autocomplete object gives users the required behavior, but the Autocomplete's exactMatch property is explicitly set to false with no way of changing it.

In this pull request, I added code to add an exactMatch configuration property to the editor object, and modified the AutoComplete object to set its exactMatch property from the configuration property's value. This allows my application to change the Autocomplete's behavior to match users' expectations.


amurph105:
I'm bumping this because this would be really useful.

akoreman:
The option to do exact matches was added as part of PR #5084, closing this PR.

PR #4975, opened 2022-10-30
Fix: update/add missing demo samples

ParkerM:
*Description of changes:*
This fixes some filename-related issues that would cause 404's to appear in the demo for certain modes, and also backfills some samples for declared modes.

To reproduce one instance of the issue, visit https://ace.c9.io/build/kitchen-sink.html and select "Fortran". The content that is loaded is 404 HTML markup instead of Fortran code.

Summary of changes:
 - `fortran.f90` changed to `fortran.f` so it'd match the 0th file extension
 - Lowercased filenames for `Haxe.hx, Jack.jack, Nix.nix` to fix mismatches
 - Added SPARQL, Turtle, and Logtalk to modelist. (These have existing samples but aren't selectable.)
 - Added samples for Django, MIPS, and Text.

The one straggler that I was unable to identify was the Log (`.log`) file type. It doesn't appear to be implemented but it is mentioned in the modelist. 


andrewnester:
Thank you, merged!

PR #5003, opened 2022-11-24
Add missing options to `EditorOptions`

dodgex:
While using ace in a small project, i realized that the options for ace/language_tool are missing in EditorOptions.

I tried to determine the types for the options in a best effort according to thier usage in
https://github.com/ajaxorg/ace/blob/v1.13.1/src/ext/language_tools.js#L152

---


andrewnester:
Thank you!

Issue #5004, opened 2022-11-25
Autocomplete/Language Tools: Potential issue with doc tooltip


dodgex:
### Describe the bug

I am working on a project using Ace and would love to utilize the Doc Tooltip in the autocompletion feature.

Unfortunately the Typescript definitions in `ace.d.ts` do not have the required definitions. As I have not found any documentation about this feature, I started to read the code to find how to implement that Tooltip for my completer. While reading the code I might have found some potential issues.

### Expected Behavior

Have proper Type Definitions and implementations to use Doc Tooltip in custom Completer.

### Current Behavior

In [autocomplete.js](https://github.com/ajaxorg/ace/blob/v1.13.1/src/autocomplete.js#L342-L346) the `updateDocTooltip()` uses [`.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) to iterate over an array of `completers` to find a completer that has a `getDocTooltip()` and returns a value for the selected item. Either, the `getDocTooltip()` (currently the only one is defined in [language_tools.js](https://github.com/ajaxorg/ace/blob/v1.13.1/src/ext/language_tools.js#L60-L67)) should return a value (`string` or `{docHTML: string}` or `{docText: string}`) to truly utilize the `.some()` or it should be a `.forEach()` instead.

### Reproduction Steps
None.

### Possible Solution

I would love to have the [`Completion` in `ace.d.ts`](https://github.com/ajaxorg/ace/blob/v1.13.1/ace.d.ts#L313-L319) include the `docHTML?: string` and/or `docText?: string` fields that could be [used directly](https://github.com/ajaxorg/ace/blob/v1.13.1/src/autocomplete.js#L347-L348).

The `getDocTooltip()` should either be defined in the [`Completer` in `ace.d.ts`](https://github.com/ajaxorg/ace/blob/v1.13.1/ace.d.ts#L909-L917) too, or in theory it could be replaced with directly defining docHTML/docText in the Completition Item in `language_tools.js`.

### Ace Version / Browser / OS / Keyboard layout

1.13.1

akoreman:
In version 1.17.0 type definitions for doc tooltips were added, see #5005 for more details.

PR #5005, opened 2022-11-25
Completions improvements

mkslanc:
*Issue #, if available:* #5004

*Description of changes:*
- Fix wrong declarations for `Completion` and `Completer`
- Add description for `Completer` structure
- Add two internal properties `range` to support range replacements and `command` for different kind of commands after insertion of completion (for now it's start new autocomplete)
- Add ability to use separate `getDocTooltip` for any completer


src/autocomplete.js
```diff
@@ -340,7 +380,7 @@ var Autocomplete = function() {
         if (!selected || !this.editor || !this.popup.isOpen)
             return this.hideDocTooltip();
         this.editor.completers.some(function(completer) {
-            if (completer.getDocTooltip)
+            if (completer.getDocTooltip && selected.completerId === completer.id)
                 doc = completer.getDocTooltip(selected);
```
dodgex, line 384: usage of a result of `getDocTooltip()` that is defined to have a void return in àce.d.ts

src/ext/language_tools.js
```diff
@@ -51,7 +53,8 @@ var snippetCompleter = {
                     caption: caption,
                     snippet: s.content,
                     meta: s.tabTrigger && !s.name ? s.tabTrigger + "\u21E5 " : "snippet",
-                    type: "snippet"
+                    type: "snippet",
+                    completerId: snippetCompleter.id
```
dodgex, line 57: while i like the idea of adding this field, i feel like it should be set in the `gatherCompletions` functions in autocomplete.js. otherwise it might be missed to be set by someone implementing it; or even setting a wrong value due to a copy/paste mistake?

dodgex:
Thank you for creating this PR. 

I added some comments with thoughts on a few of the changes.

src/ext/language_tools.js
```diff
@@ -14,8 +14,10 @@ var keyWordCompleter = {
         }
         var state = editor.session.getState(pos.row);
         var completions = session.$mode.getCompletions(state, session, pos, prefix);
+        completions = completions.map((el) => el.completerId = keyWordCompleter.id);
```
nightwing, line 17: do we have to set completerId for all completions or can do it only when needed?

src/ext/language_tools.js
```diff
@@ -51,7 +53,8 @@ var snippetCompleter = {
                     caption: caption,
                     snippet: s.content,
                     meta: s.tabTrigger && !s.name ? s.tabTrigger + "\u21E5 " : "snippet",
-                    type: "snippet"
+                    type: "snippet",
```
nightwing, line 56: type here was serving the same purpose as completerId

ace.d.ts
```diff
@@ -913,6 +916,8 @@ export namespace Ace {
       position: Point,
       prefix: string,
       callback: CompleterCallback): void;
+    getDocTooltip(item: Completion): void;
```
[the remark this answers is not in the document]
mkslanc: Agree with optional, with returning `void` it's more complicated, need to think more, because now source code and use cases contradicts each other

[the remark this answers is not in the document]
mkslanc: Changed definition to fit actual code

src/autocomplete.js
mkslanc: Changed definition of `getDocTooltip` to fit actual code

src/ext/language_tools.js
mkslanc: We would leave it for user to define this or not, with all default ace's completers with defined id's that shouldn't interfere with user's experience

src/ext/language_tools.js
mkslanc: We could define it only when needed, but I suggest to leave this for default ones (to not interfere with user defined completers)

src/ext/language_tools.js
mkslanc: Removed in new version

andredcoliveira (dismissed):
Overall looks good to me. Just a few minor type-related comments/questions.

ace.d.ts
```diff
@@ -314,11 +314,14 @@ export namespace Ace {
   }
 
   export interface Completion {
```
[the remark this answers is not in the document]
mkslanc: We need only one mandatory option, but it's either `value` or `snippet`, other options could be omitted. So I suggest something like that:
```typescript  
interface BaseCompletion {
    score?: number;
    meta?: string;
    caption?: string;
    docHTML?: string;
    docText?: string;
    completerId?: string;
  }

  export interface SnippetCompletion extends BaseCompletion {
    snippet: string;
  }

  export interface ValueCompletion extends BaseCompletion {
    value: string;
  }

  export type Completion = SnippetCompletion | ValueCompletion 
```

InspiredGuy:
hey @mkslanc could you please fix the merge conflicts?

mkslanc:
> hey @mkslanc could you please fix the merge conflicts?
Hey @InspiredGuy , sure, done!

ace.d.ts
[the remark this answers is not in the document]
andredcoliveira: I think what you suggest is good. Could you make the change?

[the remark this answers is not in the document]
mkslanc: Done!

src/autocomplete.js
```diff
@@ -9,6 +9,25 @@ var dom = require("./lib/dom");
 var snippetManager = require("./snippets").snippetManager;
 var config = require("./config");
 
+/**
+ * Completion represents a text snippet that is proposed to complete text
+ * @typedef Completion
```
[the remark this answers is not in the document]
mkslanc: Changed in new commit

andredcoliveira (approved):

PR #5008, opened 2022-12-09
Autocomplete accessibility features

aoyku:
*Issue #, if available:*

*Description of changes:*
Adds necessary roles and aria-labels for the autocomplete popup.

### DIFF
--- a/src/autocomplete/popup.js
+++ b/src/autocomplete/popup.js
@@ -9,7 +9,7 @@ var dom = require("../lib/dom");
 
 var getAriaId = function(index) {
     return `suggest-aria-id:${index}`;
-}
+};
 
 var $singleLineEditor = function(el) {
     var renderer = new Renderer(el);
@@ -126,7 +126,7 @@ var AcePopup = function(parentNode) {
         if (selected !== t.selectedNode && t.selectedNode) {
             dom.removeCssClass(t.selectedNode, "ace_selected");
             el.removeAttribute("aria-activedescendant");
-            t.selectedNode.id = "ace_line_" + t.selectedNode.id.split(":")[1];
+            t.selectedNode.removeAttribute("id");
         }
         t.selectedNode = selected;
         if (selected) {
--- a/src/autocomplete/popup.js
+++ b/src/autocomplete/popup.js
@@ -7,6 +7,10 @@ var event = require("../lib/event");
 var lang = require("../lib/lang");
 var dom = require("../lib/dom");
 
+var getAriaId = function(index) {
+    return `suggest-aria-id:${index}`;
+}
+
 var $singleLineEditor = function(el) {
     var renderer = new Renderer(el);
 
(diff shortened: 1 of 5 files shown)


src/autocomplete/popup.js
```diff
@@ -114,11 +122,21 @@ var AcePopup = function(parentNode) {
         var row = popup.getRow();
         var t = popup.renderer.$textLayer;
         var selected = t.element.childNodes[row - t.config.firstRow];
-        if (selected !== t.selectedNode && t.selectedNode)
+        var el = document.activeElement; // Active element is textarea of main editor
+        if (selected !== t.selectedNode && t.selectedNode) {
             dom.removeCssClass(t.selectedNode, "ace_selected");
+            el.removeAttribute("aria-activedescendant");
+            t.selectedNode.id = "ace_line_" + t.selectedNode.id.split(":")[1];
```
skazantsev, line 129: Why do we need this? Why can't we set `ace_line_{index}` once?

src/keyboard/textinput.js
```diff
@@ -48,7 +48,23 @@ var TextInput = function(parentNode, host) {
     // FOCUS
     // ie9 throws error if document.activeElement is accessed too soon
     try { var isFocused = document.activeElement === text; } catch(e) {}
-    
+
+    this.setAriaOptions = function(options) {
+        if (options.activeDescendant) {
+            text.setAttribute('aria-haspopup', 'true');
```
skazantsev, line 54: nit: inconsistent quoting style.

src/autocomplete/popup.js
aoyku: Because of the line 135 in the same file ->
`selected.id = ariaId;` 
While moving through the popup list, the selected element gets assigned a special id which is used as the active descendant from the main editor textarea (which is the active element). If the line is unselected, we need to set its id back to its original value.

src/layer/text.js
```diff
@@ -293,6 +293,8 @@ var Text = function(parentEl) {
                 lineEl.className = "ace_line_group";
             } else {
                 lineEl.className = "ace_line";
+                lineEl.id = "ace_line_" + row;
```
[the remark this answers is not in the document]
aoyku: Yes, right. I didn't face any issues probably because I am setting another id (aria-id) to the selected line to be able to point to this line as an active descendant from the active element. 

What I could do is number each editor  (as you suggested), specify each line's id with which editor it belongs and its row number.

src/autocomplete/popup.js
aoyku: I removed the second "ace_line_{index}" id. We now only have the id that I set for aria.

andrewnester (approved):

skazantsev (commented):
In the current implementation we set an id of the selected element in the autocomplete and remove the id once the element is not selected/focussed. Why we need to do all of this? Can't we generate id once e.g. ace_line_{index} and set `aria-activedescendant` when an item is selected? Ofc, we need to make sure that ids do not clash.

PR #5011, opened 2022-12-12
Update ace.d.ts typings for navigate

sadnub:
*Description of changes:*

The typings navigateUp/Down/LeftRight are missing the times parameter. This just adds them to match the function definition.

This will show errors in an IDE:
`editor.navigateUp(3)`

### DIFF
--- a/ace.d.ts
+++ b/ace.d.ts
@@ -883,10 +883,10 @@ export namespace Ace {
     jumpToMatching(select: boolean, expand: boolean): void;
     gotoLine(lineNumber: number, column: number, animate: boolean): void;
     navigateTo(row: number, column: number): void;
-    navigateUp(times: number): void;
-    navigateDown(times: number): void;
-    navigateLeft(times: number): void;
-    navigateRight(times: number): void;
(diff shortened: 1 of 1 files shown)


ace.d.ts
```diff
@@ -883,10 +883,10 @@ export namespace Ace {
     jumpToMatching(select: boolean, expand: boolean): void;
     gotoLine(lineNumber: number, column: number, animate: boolean): void;
     navigateTo(row: number, column: number): void;
-    navigateUp(): void;
-    navigateDown(): void;
-    navigateLeft(): void;
-    navigateRight(): void;
+    navigateUp(times: number): void;
```
andrewnester, line 886: `times` parameter is optional for all 4 methods, could you please update it to `times?: number`?

andrewnester (changes requested):

ace.d.ts
sadnub: Just updated. Thanks!

andrewnester (approved):

andrewnester:
Thank you!

PR #5073, opened 2023-03-01
Fix vim paste

whazor:
*Issue #, if available:* #2680

*Description of changes:* 
Fixed VIM paste with new way of how browser allow clipboard. This will show a browser pasting allow/deny permission window, which should be expected for end-user when pasting.

Set default register back to '+', this allows browser pasting again.

`navigator.permissions.query` has broad [browser support](https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query#browser_compatibility).

Firefox does not allow usage of `clipboard-write`, and will fall back to normal pasting, no browser permission window shown.


whazor:
Test fails `Yy_blockwise`: https://github.com/ajaxorg/ace/blob/bd3573ce0ecb095e962e78b4ab2c845cfa40fdcb/src/keyboard/vim_test.js#L1762

After the first 5 lines, the results are widely different from vim, vim main branch, vim my branch.

Main branch:
```
123456
1233456456
123456
123456
```

Vim:
```
123456
1233456456
123
123
```

My branch:
```
112345623456
112334562345645623456
```

PR #5084, opened 2023-03-06
feat: inline autocompletion

azmkercso:
*Issue #, if available:* N/A

*Description of changes:*
Inline autocompletion feature.
Minor fixes to ghost text display.
Snippet manager preview text.

The popup-based autocompletion is augmented with optional inline-code rendering. This feature is turned off by default, so no unexpected ghost text appears for users in the new version. It can be turned on with the `inlineEnabled` boolean flag in the `Autocomplete` widget.
This autocompletion displays the inline completions on a best effort basis (if the completion's prefix allows it), as the autocomplete settings might not be fully compatible with inline completion (eg: matches based on caption not prefix might cannot be rendered inline).
If the inline completions are enabled, the popup position is dynamically updated whenever it would overlay a multi-line inline completion suggestion.

An other external widget was created which can be used for inline-only completion. Here, all the completions are guaranteed to be compatible with inline rendering, and the commands are set up in a way to promote more horizontal (left-right, rather than up-down) controls for autocompletion. This widget could be used as is, to allow both style of completions, or could be used as a reference implementation for integrators.

This external inline autocompletion widget works similarly to the previous popup-based autocompletion with a few differences:
- It displays a slightly different set of completions (strict prefix match with `exactMatch`, and does not try to match on `caption` value)
- It always renders a ghost text completion directly into the editor
- It optionally shows a command bar tooltip with the available commands for easy navigation. These commands are dynamically updated, and clickable.
- Editor scrolling with mouse does not cancel the autocompletion (so long snippets can be also inspected)

Keybindings:
- Manual trigger: `Alt/Option-C`
- Previous suggestion: `Alt/Option-[`
- Next suggestion: `Alt/Option-]`
- Accept suggestion: `Tab` or `Ctrl/Cmd-Right`
- Close suggestions: `Esc`

The popup and the inline style autocompletions are mutually exclusive, only one is attached at a time (they share the same `editor.completer` property). However, they can seamlessly switch when both of them are configured to different keybindings. 

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -42,40 +42,49 @@ var Autocomplete = function() {
 };
 
 (function() {
-    this.$onPopupChange = function(hide) {
-        if (hide) {
-            if (this.inlineEnabled) {
-                this.inlineRenderer.show(this.editor, null, prefix);    
-            }
-            this.hideDocTooltip();
-        } else {
-            this.tooltipTimer.call(null, null);
-            if (this.inlineEnabled) {
-                var completion = hide ? null : this.popup.getData(this.popup.getRow());
-                var prefix = util.getCompletionPrefix(this.editor);
-                this.inlineRenderer.show(this.editor, completion, prefix);
-                this.$updatePopupPosition();
-            }
-        }
-    };
 
     this.$init = function() {
         this.popup = new AcePopup(document.body || document.documentElement);
-        this.inlineRenderer = new AceInline();
         this.popup.on("click", function(e) {
             this.insertMatch();
             e.stop();
         }.bind(this));
         this.popup.focus = this.editor.focus.bind(this.editor);
         this.popup.on("show", this.$onPopupChange.bind(this));
-        this.popup.on("hide", this.$onPopupChange.bind(this, true));
+        this.popup.on("hide", this.$onHidePopup.bind(this));
         this.popup.on("select", this.$onPopupChange.bind(this));
         this.popup.on("changeHoverMarker", this.tooltipTimer.bind(null, null));
         return this.popup;
     };
 
+    this.$initInline = function() {
+        if (!this.inlineEnabled || this.inlineRenderer)
+            return;
+        this.inlineRenderer = new AceInline();
+        return this.inlineRenderer;
+    }
+
     this.getPopup = function() {
         return this.popup || this.$init();
+    };
+
+    this.$onHidePopup = function() {
+        if (this.inlineRenderer) {
+            this.inlineRenderer.hide();
+        }
+        this.hideDocTooltip();
+    }
+
+    this.$onPopupChange = function(hide) {
+        this.tooltipTimer.call(null, null);
+        if (this.inlineRenderer && this.inlineEnabled) {
+            var completion = hide ? null : this.popup.getData(this.popup.getRow());
+            var prefix = util.getCompletionPrefix(this.editor);
+            if (!this.inlineRenderer.show(this.editor, completion, prefix)) {
+                this.inlineRenderer.hide();
+            }
+            this.$updatePopupPosition();
+        }
     };
 
     this.$updatePopupPosition = function() {
@@ -103,6 +112,9 @@ var Autocomplete = function() {
     this.openPopup = function(editor, prefix, keepPopupPosition) {
         if (!this.popup)
             this.$init();
+
+        if (this.inlineEnabled && !this.inlineRenderer)
+            this.$initInline();
 
         this.popup.autoSelect = this.autoSelect;
 
@@ -135,11 +147,13 @@ var Autocomplete = function() {
      * Detaches all elements from the editor, and cleans up the data for the session
      */
     this.detach = function() {
-        this.editor.keyBinding.removeKeyboardHandler(this.keyboardHandler);
-        this.editor.off("changeSelection", this.changeListener);
-        this.editor.off("blur", this.blurListener);
-        this.editor.off("mousedown", this.mousedownListener);
-        this.editor.off("mousewheel", this.mousewheelListener);
+        if (this.editor) {
+            this.editor.keyBinding.removeKeyboardHandler(this.keyboardHandler);
+            this.editor.off("changeSelection", this.changeListener);
+            this.editor.off("blur", this.blurListener);
+            this.editor.off("mousedown", this.mousedownListener);
+            this.editor.off("mousewheel", this.mousewheelListener);
+        }
         this.changeTimer.cancel();
         this.hideDocTooltip();
 
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -2,6 +2,7 @@
 
 var HashHandler = require("./keyboard/hash_handler").HashHandler;
 var AcePopup = require("./autocomplete/popup").AcePopup;
+var AceInline = require("./autocomplete/inline").AceInline;
 var getAriaId = require("./autocomplete/popup").getAriaId;
 var util = require("./autocomplete/util");
 var lang = require("./lib/lang");
@@ -9,11 +10,22 @@ var dom = require("./lib/dom");
 var snippetManager = require("./snippets").snippetManager;
 var config = require("./config");
 
+var destroyCompleter = function(e, editor) {
+    editor.completer && editor.completer.destroy();
+};
+
+
+/**
+ * This object controls the autocompletion components and their lifecycle.
+ * There is an autocompletion popup, an optional inline ghost text renderer and a docuent tooltip popup inside.
+ * @class
+ */
+
 var Autocomplete = function() {
     this.autoInsert = false;
     this.autoSelect = true;
     this.exactMatch = false;
-    this.gatherCompletionsId = 0;
+    this.inlineEnabled = false;
     this.keyboardHandler = new HashHandler();
     this.keyboardHandler.bindKeys(this.commands);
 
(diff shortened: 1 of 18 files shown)


src/autocomplete/inline_tooltip.js
```diff
@@ -0,0 +1,168 @@
...
+ * @param {HTMLElement} parentElement
+ * @param {Record<string,TooltipCommand>} commands
+ * @constructor
+ */
+
+var Constants = Object.freeze({
+    entryClassName: 'inline_autocomplete_tooltip_entry',
+    buttonClassName: 'inline_autocomplete_tooltip_button',
+    tooltipClassName: 'ace_tooltip ace_inline_autocomplete_tooltip',
+    tooltipId: 'inline_autocomplete_tooltip'
```
nightwing, line 21: better to use `var TOOLTIP_ID =` instead of the weird trick with freeze.

src/autocomplete/inline_tooltip.js
```diff
@@ -0,0 +1,168 @@
...
+        parentElement.appendChild(el);
+    }
+    el.id = Constants.tooltipId;
+    el.style['pointer-events'] = 'auto';
+    el.className = Constants.tooltipClassName;
+    this.commands = commands;
+    this.buttons = {};
+    this.eventListeners = {};
+
+    Object.keys(commands).forEach(function (key) {
```
nightwing, line 38: using buildDom from lib/dom may make this code simpler.

src/autocomplete/inline_tooltip.js
```diff
@@ -0,0 +1,168 @@
...
+dom.importCssString(`
+.ace_inline_autocomplete_tooltip {
+    display: inline-block;
+}
+.inline_autocomplete_tooltip_entry {
+    display: inline-block;
+    padding: 0 10px;
+}
+
+.inline_autocomplete_tooltip_button:disabled {
```
nightwing, line 163: buttons need some hover style and maybe cursor pointer.

src/inline_autocomplete.js
```diff
@@ -0,0 +1,356 @@
...
+            if (completion.snippet)
+                snippetManager.insertSnippet(this.editor, completion.snippet);
+            else
+                this.editor.execCommand("insertstring", completion.value || completion);
+        }
+        this.editor.endOperation();
+        this.detach();
+    };
+
+    this.gatherCompletions = function(callback) {
```
nightwing, line 200: Is it not possible to reuse this function instead of duplicating it?

nightwing (commented):
Looks very good, but overall the choice of replacing the existing autocompletion instead of augmenting it, is unexpected for me. Some users may want the behavior of chrome devtools where it both shows the completion inline and a normal popup.

Also i think it would be better to keep this in a separate extension instead of adding to language_tools, because the vast majority of users would not be using this.

And another question, what should happen if the line has an autopaired brace inserted? E.g. if user writes `if (` editor inserts closing `)` and completer returns a result for the whole if block?

src/autocomplete.js
```diff
@@ -444,51 +447,75 @@ var Autocomplete = function() {
...
-var FilteredList = function(array, filterText) {
+var FilteredList = function(array, filterText, prefix) {
     this.all = array;
     this.filtered = array;
     this.filterText = filterText || "";
+    this.prefix = prefix || "";
     this.exactMatch = false;
 };
 (function(){
+    this.setPrefix = function(str) {
```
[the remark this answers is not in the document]
azmkercso: Yeah, I overlooked that one, will fix by using that.

src/autocomplete/inline.js
```diff
@@ -0,0 +1,160 @@
...
+        }.bind(this));
+        if (currentCompletionIsAvailable) {
+            this.$index = newIndex;
+        } else {
+            this.$index = newIndex === -1 ? -1 : 0;
+        }
+        this.renderCompletion();
+    };
+
+    Object.defineProperties(this, {
```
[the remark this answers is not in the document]
azmkercso: Okay, I'll adjust it to match the style of the other APIs

azmkercso:
>Looks very good, but overall the choice of replacing the existing autocompletion instead of augmenting it, is unexpected for me. Some users may want the behavior of chrome devtools where it both shows the completion inline and a normal popup.
Yes, that's a very good point, I am working on a new revision to reuse the current autocompletion instead of a new one and show the popup with inline text. However inline-text only mode requires slightly different navigation and rules, so I think the two different types of completer makes sense (inline-only, and popup with optional inline text).

>Also i think it would be better to keep this in a separate extension instead of adding to language_tools, because the vast majority of users would not be using this.
Sure, removing it from there.

azmkercso:
>And another question, what should happen if the line has an autopaired brace inserted? E.g. if user writes if ( editor inserts closing ) and completer returns a result for the whole if block?
Currently when I tested, no autopaired brace was inserted while the autocomplete was active. However the behaviour is not handled explicitly at the moment.
It could either:
- abort the autocompletion, or
- implement the `deleteSuffix` option, or
- treat only the left prefix of the completion for the possible suggestions, but also replace the additional `)` on insert if the first parantheses or the suggested inline text matches the inserted pair (in this case `)`).

azmkercso:
Update: enabled optional inline completion, and moved the inline-only autocompletion to a separate extension.
Details are in the updated description

src/inline_autocomplete.js
```diff
@@ -0,0 +1,356 @@
...
+InlineAutocomplete.startCommand = {
+    name: "startInlineAutocomplete",
+    exec: function(editor, options) {
+        var completer = InlineAutocomplete.for(editor);
+        if (completer instanceof InlineAutocomplete) {
+            completer.showInlineCompletions(editor, options);
+        }
+        return completer;
+    },
+    bindKey: "Alt-C"
```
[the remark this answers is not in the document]
azmkercso: I use mac, and it worked for me with only this, but I agree it is better if it is specified explicitly, rather than relying on an implicit conversion.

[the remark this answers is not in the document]
nightwing: I meant that with option-c we are very likely to get complaints about some character input on international keyboards not working. At least that's what have happened with our other attempts to use option- on mac.

But since this feature is not enabled by default, i am ok with keeping this as is if that is the desired binding for cloud9

andrewnester (approved):

src/autocomplete.js
```diff
@@ -30,16 +42,34 @@ var Autocomplete = function() {
...
+                var prefix = util.getCompletionPrefix(this.editor);
+                this.inlineRenderer.show(this.editor, completion, prefix);
+                this.$updatePopupPosition();
+            }
+        }
+    };
 
     this.$init = function() {
         this.popup = new AcePopup(document.body || document.documentElement);
+        this.inlineRenderer = new AceInline();
```
nightwing, line 64: Do we always need this or only when inlineEnabled is true?

src/autocomplete.js
```diff
@@ -30,16 +42,34 @@ var Autocomplete = function() {
...
+    this.$onPopupChange = function(hide) {
+        if (hide) {
+            if (this.inlineEnabled) {
+                this.inlineRenderer.show(this.editor, null, prefix);    
+            }
+            this.hideDocTooltip();
+        } else {
+            this.tooltipTimer.call(null, null);
+            if (this.inlineEnabled) {
+                var completion = hide ? null : this.popup.getData(this.popup.getRow());
```
nightwing, line 54: I think `hide` never can be true here, maybe worth to split into two functions to make it easier to follow?

nightwing:
- when cursor is at the top of the page toolbar is not visible
- i would expect alt-c the second time to act as next command
- the () around alt-[ make it look like alt-D
- when pressing alt-c at the start empty line, toolbar behaves as if it is cycling through suggestions but editor doesn't show anything

the last issue seems to be caused by a bug in addToken which should do `if (column == null || !tokens.lenght) {` at https://github.com/ajaxorg/ace/blob/e0d4ea9bfb00e0ea538822268769b050347f19c1/src/virtual_renderer.js#L1665

src/autocomplete.js
azmkercso: It is only needed when inlineEnabled is true, I'll make sure it's not created needlessly.

src/autocomplete.js
azmkercso: Sure, I'll split it

src/ext/inline_autocomplete_test.js
```diff
@@ -0,0 +1,286 @@
+
+if (typeof process !== "undefined") {
+    require("amd-loader");
+    require("../test/mockdom");
+}
+
+"use strict";
+
+var ace = require("../ace");
```
andredcoliveira, line 9: This variable seems to be unused

src/ext/inline_autocomplete_test.js
```diff
@@ -0,0 +1,286 @@
...
+    setUp: function(done) {
+        var el = document.createElement("div");
+        el.style.left = "20px";
+        el.style.top = "30px";
+        el.style.width = "500px";
+        el.style.height = "500px";
+        document.body.appendChild(el);
+        var renderer = new VirtualRenderer(el);
+        var session = new EditSession("");
+        editor = new Editor(renderer, session);
```
andredcoliveira, line 75: Should we test more than one editor too?

andredcoliveira (commented):
Left some comments. Would it be possible to add this to the _kitchen sink_ so we can play around with it too?

src/ext/inline_autocomplete_test.js
azmkercso: I adjusted this class to be editor scoped.

src/ext/inline_autocomplete.js
```diff
@@ -0,0 +1,611 @@
...
+     * Displays the clickable command bar tooltip
+     * @param {Record<string, TooltipCommand>} commands
+     */
+    this.show = function(editor) {
+        this.detach();
+
+        this.htmlElement.style.display = '';
+        this.htmlElement.addEventListener('mousedown', captureMousedown.bind(this));
+        
+        this.editor = editor;
```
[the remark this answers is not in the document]
azmkercso: Adjusted it to be editor scoped

azmkercso:
>Would it be possible to add this to the kitchen sink so we can play around with it too?

I added it to the kitchen sink so it can be tested. `Alt/Option-C` triggers the inline autocompletion, and the usual `Ctrl-Space` triggers the popup based autocompletion. Only one can be active at a time, the other one is destroyed if the other autocompleter is triggered.

I also added an option which overrides the `inlineEnabled` option for Autocomplete which is disabled by default (mainly to keep the current behaviour after the update). With this, the popup based autocompletion also renders an inline preview (if possible).

andrewnester (approved):

andrewnester (approved):

andrewnester (approved):

andredcoliveira (approved):
Looks good to me. Nice set of changes and the addition to the kitchen sink was very much appreciated :).

There is only one detail/edge case that I'd like to leave a note here of:
* when showing inline autocompletion suggestions that contain multiple suggested lines and they go beyond the bottom of the editor, specifically when using the default autocompletion interaction (e.g.: `⌥` + `Space` for MacOS, not `⌥` + `C`), the blur behaviour of the auto-completion popup that is to stop showing the suggestion makes it so it's not possible to scroll lower in the editor and see all suggested lines.
  * I don't find this to be a blocker to this PR

Issue #5090, opened 2023-03-10
Error in mode-latex getType


kurtslagle:
When using `mode-latex`, a console error appears saying `Cannot read properties of null (reading 'type')`, coming from https://github.com/ajaxorg/ace-builds/blob/master/src-noconflict/mode-latex.js#L142

It seems like whenever there isn't any content after the cursor, typing `\begin` or `\end` to begin or end a new environment causes a console error to show up.

It's simple to reproduce in the kitchen sink: https://ajaxorg.github.io/ace-builds/kitchen-sink.html
Just set the mode to `LaTeX` and with this document content

```latex
\usepackage{amsmath}
\title{\LaTeX}
\date{}
<start typing here>
```

begin typing `\begin{document}`. Once you finish typing `begin`, the console error will show up.

This doesn't appear to happen if there is any text after where you would be typing `\begin{...`, like this

```latex
<start typing here>
some other stuff
```
typing `\begin` does not make the error appear.

andrewnester:
Released in 1.16.0

kurtslagle:
Thank you for fixing that so quickly!

PR #5093, opened 2023-03-15
Fix bug with missing token in latex folding

mkslanc:
*Issue #, if available:* #5090

*Description of changes:*


InspiredGuy:
@mkslanc the fix looks good, but the coverage check for the patch failed. Could you please add a simple test for the function you modified, so this bug does not reappear again? I see we have tests for some other modes already.

mkslanc:
> some

Sure, added test similar to other folding tests

InspiredGuy (approved):

PR #5107, opened 2023-03-26
Add option to use SVG gutter icons

akoreman:
Adds an option (`useSvgGutterIcons`) to use SVG gutter icons (only annotation icons for now). By default `useSvgGutterIcons` is false, after setting `useSvgGutterIcons = true` the gutter annotations will be drawn using a CSS mask on top of a background color. This allows to set the stroke color of the SVG icons in themes.

Screenshot 2023-04-03 at 21 36 05

When `useSvgGutterIcons = false` we return to the current behavior: 

Screenshot 2023-03-25 at 23 00 10


nightwing:
On windows, for monitors without high resolution the image does not have proper antialiasing, and looks rather jagged image
Maybe we should try to use font, or use css filter to change svg color (though i am not sure about the performance of the later case)

akoreman:
@nightwing Could you share what resolution and browser you're using (I'm looking at Chrome on Windows 11 at a resolution of 800x600px and I don't see the jagged edges).

Using CSS filters can make it hard to set the color to a particular RGB/HEX value, you'd have to lookup the correct Hue value. I don't think switching to using CSS filters would affect how the UA applies AA here?

nightwing:
I am seeing the above on chrome windows 10 1680x1050 display

akoreman:
Made some tweaks to the SVGs which could help here.

InspiredGuy (approved):

PR #5117, opened 2023-04-06
feat: summary of annotations in folded lines

akoreman:
Currently, when there are gutter annotations for lines which are folded away, the gutter annotations are invisible. This change adds a (optional) one-line summary of the annotations in the folded code:

Screenshot 2023-04-16 231107

Screenshot 2023-04-20 at 10 46 01

Added as an option to not break existing users who have custom gutter annotation icons in their themes (they would see their custom icons for 'normal' annotations and the default icons for annotations in folds). 


nightwing:
The change of anchor position was accidental in the last pr, but maybe we should keep it aligned to the icon? I think the first one looks better, and it is worth to consider keeping it as is.

akoreman:
The main problem I noticed was the inconsistency between the anchoring for SVG and PNG icon gutter tooltips so I changed it to be the old behavior for both. But I'm also okay with changing it so we will have the first behavior for both, agree that the first one looks a bit more coherent.

src/css/editor.css.js
```diff
@@ -115,29 +115,34 @@ module.exports = `
...
     background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAMAAAAoLQ9TAAABOFBMVEX/////////QRswFAb/Ui4wFAYwFAYwFAaWGAfDRymzOSH/PxswFAb/SiUwFAYwFAbUPRvjQiDllog5HhHdRybsTi3/Tyv9Tir+Syj/UC3////XurebMBIwFAb/RSHbPx/gUzfdwL3kzMivKBAwFAbbvbnhPx66NhowFAYwFAaZJg8wFAaxKBDZurf/RB6mMxb/SCMwFAYwFAbxQB3+RB4wFAb/Qhy4Oh+4QifbNRcwFAYwFAYwFAb/QRzdNhgwFAYwFAbav7v/Uy7oaE68MBK5LxLewr/r2NXewLswFAaxJw4wFAbkPRy2PyYwFAaxKhLm1tMwFAazPiQwFAaUGAb/QBrfOx3bvrv/VC/maE4wFAbRPBq6MRO8Qynew8Dp2tjfwb0wFAbx6eju5+by6uns4uH9/f36+vr/GkHjAAAAYnRSTlMAGt+64rnWu/bo8eAA4InH3+DwoN7j4eLi4xP99Nfg4+b+/u9B/eDs1MD1mO7+4PHg2MXa347g7vDizMLN4eG+Pv7i5evs/v79yu7S3/DV7/498Yv24eH+4ufQ3Ozu/v7+y13sRqwAAADLSURBVHjaZc/XDsFgGIBhtDrshlitmk2IrbHFqL2pvXf/+78DPokj7+Fz9qpU/9UXJIlhmPaTaQ6QPaz0mm+5gwkgovcV6GZzd5JtCQwgsxoHOvJO15kleRLAnMgHFIESUEPmawB9ngmelTtipwwfASilxOLyiV5UVUyVAfbG0cCPHig+GBkzAENHS0AstVF6bacZIOzgLmxsHbt2OecNgJC83JERmePUYq8ARGkJx6XtFsdddBQgZE2nPR6CICZhawjA4Fb/chv+399kfR+MMMDGOQAAAABJRU5ErkJggg==");
     background-repeat: no-repeat;
     background-position: 2px center;
 }
 
-.ace_gutter-cell.ace_warning, .ace_icon.ace_warning {
+.ace_gutter-cell.ace_warning, .ace_gutter-tooltip .ace_icon.ace_warning {
```
nightwing, line 134: Why is this change needed? We may want to use the icon in other places too, e.g. in documentation tooltip.

src/css/editor.css.js
```diff
@@ -115,29 +115,34 @@ module.exports = `
...
     float: left;
 }
 
-.ace_gutter-cell.ace_error, .ace_icon.ace_error {
+.ace_gutter-cell .ace_icon {
+    margin-left: -18px;
+    float: left;
+}
+
+.ace_gutter-cell.ace_error, .ace_gutter-tooltip .ace_icon.ace_error {
```
nightwing, line 128: I think this change will break cloud9, because there we use custom gutter cell backgrounds for errors and breakpoints. With this both will be visible. 

If the goal is simply to have a node for getting client rect, maybe we can use client rect from element and offset by `editor.renderer.$gutterLayer.$padding.left`?

akoreman: Let's try to not break c9 🙂 could you explain how this change could affect c9? 

I assume they use selectors like `.ace_gutter-cell.ace_error` for their custom backgrounds and nothing changes for these selectors. The only change here is that the background set for `.ace_icon.ace_error` is now scoped to `.ace_gutter-tooltip` (to avoid the background image showing up twice).

nightwing: My bad, i thought this changes ace to set background on `ace_icon` instead of `ace_gutter-cell`, which would break the css override in cloud9.

I still think using the value from padding instead of adding a completely empty element would be better, but this may be ok too.

akoreman:
I'm working on incorporating this CR into a slightly larger improvement of the gutter tooltip. Setting PR to draft for now.

src/mouse/default_gutter_handler.js
```diff
@@ -94,7 +119,7 @@ function GutterHandler(mouseHandler) {
         if (mouseHandler.$tooltipFollowsMouse) {
             moveTooltip(mouseEvent);
         } else {
-            var gutterElement = mouseEvent.domEvent.target;
+            var gutterElement = gutter.$lines.cells[row].element.children[1];
```
InspiredGuy, line 122: what is `gutter.$lines.cells[row].element.children[1]`? is there a better way to get that element instead of relative positioning of element among other children?

src/mouse/default_gutter_handler.js
```diff
@@ -34,12 +34,60 @@ function GutterHandler(mouseHandler) {
...
+            annotation = {text: [...annotationsInRow.text], type: [...annotationsInRow.type]};
+        else
+            annotation = {text: [], type: []};
+
+        // If the tooltip is for a row which has a closed fold, check whether there are
+        // annotations in the folded lines. If so, add a summary to the list of annotations.
+        var fold = gutter.session.getNextFoldLine(row);
+        if (fold && gutter.$showFoldedAnnotations){
+            var annotationsInFold = {error: [], warning: [], info: []};
+            var typeMostSevereAnnotationInFold;
```
InspiredGuy, line 60: `typeMostSevereAnnotationInFold` -> `typeOfMostSevereAnnotationInFold` or `mostSevereAnnotationInFoldType`?

src/mouse/default_gutter_handler.js
```diff
@@ -105,14 +130,28 @@ function GutterHandler(mouseHandler) {
...
             editor._signal("hideGutterTooltip", tooltip);
             editor.off("mousewheel", hideTooltip);
         }
     }
 
+    function annotationsToSummaryString(annotations) {
+        var isMoreThanOneAnnotationType = false;
+        var summaryString = "";
+        for (var i = 0; i < 3; i++){
+            var annotationType = ['error', 'warning', 'info'][i];
```
InspiredGuy, line 145: we can simplify and get rid of magic number 3 here. we can use es6 since it's all over the codebase now
```js
const annotationTypes = ['error', 'warning', 'info'];
for (const annotationType of annotationTypes) {...}
```
and if `annotationTypes` are also used/available somewhere else, would be great to have them specified in a single place

src/mouse/default_gutter_handler.js
```diff
@@ -105,14 +130,28 @@ function GutterHandler(mouseHandler) {
...
+            if (annotations[annotationType].length > 0){
+                var label = annotations[annotationType].length === 1 ? annotationLabels[annotationType].singular : annotationLabels[annotationType].plural;
+                summaryString += `${isMoreThanOneAnnotationType ? ', ' : ''}${annotations[annotationType].length} ${label}`;
+                isMoreThanOneAnnotationType = true;
+            }
```
InspiredGuy, line 150: nit: maybe early exit to have one less level of indentation?
```js
if (!annotations[annotationType].length) continue;
...
```

src/mouse/default_gutter_handler.js
```diff
@@ -105,14 +130,28 @@ function GutterHandler(mouseHandler) {
...
             tooltip.hide();
-            tooltipAnnotation = null;
+            tooltipContent = null;
             editor._signal("hideGutterTooltip", tooltip);
             editor.off("mousewheel", hideTooltip);
         }
     }
 
+    function annotationsToSummaryString(annotations) {
+        var isMoreThanOneAnnotationType = false;
```
InspiredGuy, line 142: I think this one is not needed at all if you use [Array.join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join). this function could look like:
```js
function annotationsToSummaryString(annotations) {
    const summary = [];
    const annotationTypes = ['error', 'warning', 'info'];
    for (const annotationType of annotationTypes) {
        if (!annotations[annotationType].length) continue;
        const label = annotations[annotationType].length === 1 ? annotationLabels[annotationType].singular : annotationLabels[annotationType].plural;
        summary.push(`${annotations[annotationType].length} ${label}`);
    }
    return summary.join(", ");
}
```

InspiredGuy (commented):
looks good, but left some comments: mostly about code style and nitpicks

src/mouse/default_gutter_handler.js
akoreman: this would get the same result:
```
var gutterElement = gutter.$lines.cells[row].element.querySelector("[class*=ace_icon]");
```

src/mouse/default_gutter_handler.js
akoreman: good call

src/ext/options.js
```diff
@@ -198,6 +198,9 @@ var optionGroups = {
         "Use SVG gutter icons": {
             path: "useSvgGutterIcons"
         },
+        "Show annotations folded lines": {
```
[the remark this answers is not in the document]
akoreman: Reason I didn't add that is because that pushes `lines` to a new line but I agree that `Show annotations for folded lines` is clearer to read. Not sure what's the right choice here 😄 

Screenshot 2023-04-19 at 13 44 45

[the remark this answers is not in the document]
InspiredGuy: it's a demo, it doesn't matter how many lines are there in a label. if you want to keep it in a single line, just name it `Annotations for folded lines`, checkbox is informative enough to replace the `show` word

src/mouse/default_gutter_handler.js
```diff
@@ -34,12 +34,60 @@ function GutterHandler(mouseHandler) {
...
 
     function showTooltip() {
         var row = mouseEvent.getDocumentPosition().row;
-        var annotation = gutter.$annotations[row];
-        if (!annotation)
+        var annotationsInRow = gutter.$annotations[row];
+        var annotation;
+
+        if (annotationsInRow)
+            annotation = {text: [...annotationsInRow.text], type: [...annotationsInRow.type]};
```
nightwing, line 51: let's not use code constructs that cannot be efficiently compiled for old browsers, also why do we need to copy the array here?

src/mouse/default_gutter_handler.js
```diff
@@ -34,12 +34,60 @@ function GutterHandler(mouseHandler) {
...
+        else
+            annotation = {text: [], type: []};
+
+        // If the tooltip is for a row which has a closed fold, check whether there are
+        // annotations in the folded lines. If so, add a summary to the list of annotations.
+        var fold = gutter.session.getNextFoldLine(row);
```
nightwing, line 57: use getFoldLine instead of nextFoldline to not display annottations on rows without icon or fold

src/mouse/default_gutter_handler.js
akoreman: I want to add an element to the array used to construct the tooltip so I want to copy the array by value. Switched to a different implementation for that to not use spread.

src/layer/gutter.js
```diff
@@ -338,8 +322,29 @@ class Gutter{
...
+            if (c == "start" && row == foldStart && row < fold.end.row){
                 className += " ace_closed";
+                var foldAnnotationClass;
+                var annotationInFold = false;
+
+                for (var i = row + 1; i <= fold.end.row; i++){
+                    if (!this.$annotations[i])
+                        continue;
+
+                    if (this.$annotations[i].className === " ace_error"){
```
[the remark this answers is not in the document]
InspiredGuy: This is not something which was added in this PR, so we could probably do it in a followup.

InspiredGuy (approved):

Issue #5145, opened 2023-04-28
Tooltips for collapse/expand code blocks are missing


LidiyaGeorgieva:
### Describe the bug

Ace editor don't have tooltips for expand and collapse buttons in the editor field, which is not conform to this Accessibility requirement: 
All semantic non-text content (e.g. graphics, videos) has to have a text alternative.
Accessibility specification: https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html
![image](https://user-images.githubusercontent.com/7242508/235063125-735c10c2-123f-404c-9b1f-02188ddd90a5.png)

### Expected Behavior

Buttons for expanding and collapsing code blocks to have tooltips.

### Current Behavior

They don't have tooltips.

### Reproduction Steps

Open https://ace.c9.io/ and hover over the buttons

### Ace Version / Browser / OS / Keyboard layout

Latest version

akoreman:
[Ace 1.20.0](https://github.com/ajaxorg/ace/releases/tag/v1.20.0) has been released which includes this change.

PR #5146, opened 2023-04-28
feat: Add gutter controls to keyboard accessibility mode

akoreman:
*Issue: #5125,  #5145*

Adds operating the fold controls and annotations in the gutter to the keyboard accessibility mode. When tabbing through the page, you can focus on the gutter div. When pressing enter, you enter the gutter and can use the arrow keys to navigate through the controls in the gutter. Press enter to interact with the controls. Press escape to set focus back to the gutter div and be able to TAB through the page again.

Only iterates over gutter controls in current viewbox to prevent potentially scrolling the cursor outside of the current view (which can cause a11y problems).

https://user-images.githubusercontent.com/34573235/235914151-b3c64385-57ba-4f5a-9237-633042301037.mov

Additionally adds title attribute to folding buttons:

Screenshot 2023-05-01 200955

Screenshot 2023-05-01 201017

Changes to `src/mouse/default_gutter_handler.js` are moving things around to to it easier to use `GutterTooltip` outside of `GutterMouseHandler`.


src/editor.js
```diff
@@ -2874,34 +2875,68 @@ config.defineOptions(Editor.prototype, "editor", {
...
+                this.renderer.content.setAttribute("aria-roledescription", "");
+                this.renderer.content.classList.remove(this.keyboardFocusClassName);
                 this.renderer.content.setAttribute("aria-label", "");
             
                 this.renderer.content.removeEventListener("keyup", focusOnEnterKeyup.bind(this));
                 this.commands.removeCommand(blurCommand);
+
+                this.renderer.$gutter.setAttribute("tabindex", -1);
+                this.renderer.$gutter.setAttribute("aria-hidden", true);
+                this.renderer.content.setAttribute("role", "");
```
oykuyilmaz, line 2933: What is a best approach for removing the aria attributes, is it set them to empty value or remove them completely. I guess I followed a different approach and asking to be consistent.

src/keyboard/textinput.js
```diff
@@ -61,9 +61,21 @@ var TextInput = function(parentNode, host) {
         }
         if (options.role) {
             text.setAttribute("role", options.role);
-        }
+        }     
+    };
+    this.setAriaLabel = function() {
```
oykuyilmaz, line 66: can we maybe merge two functions here: `setAriaOptions` and `setAriaLabel`? They seem serving similar purposes.

src/editor.js
akoreman: I don't think there are differences between how assistive tech interpret it but it probably makes more sense to just remove the attributes. I'll update the PR.

src/keyboard/textinput.js
akoreman: If we merge the functions, we'd need to make sure that whenever we call `setAriaLabel` we pass the same `options` as the last time we called `setAriaOptions` which I think would be a hassle given that e.g. autocomplete extensions also call this function.

src/keyboard/gutter_handler.js
```diff
@@ -0,0 +1,417 @@
...
+                }
+
+                if (nearestFoldIndex !== null && nearestAnnotationIndex === null){
+                    this.activeRowIndex = nearestFoldIndex;
+                    this.activeLane = "fold";
+                    this.$focusFoldWidget(this.activeRowIndex);
+                    return;
+                }
+
+                if (Math.abs(nearestAnnotationIndex - index) < Math.abs(nearestFoldIndex - index)){
```
oykuyilmaz, line 73: if they are the same, `nearestFoldIndex` will get the keyboard focus, isn't it?

oykuyilmaz (approved):

PR #5147, opened 2023-04-30
Special trigger characters to start autocomplete

mkslanc:
*Issue #, if available:*

*Description of changes:*


src/autocomplete/util.js
```diff
@@ -54,3 +54,15 @@ exports.getCompletionPrefix = function (editor) {
     }.bind(this));
     return prefix || this.retrievePrecedingIdentifier(line, pos.column);
 };
+
+exports.triggerAutocomplete = function (editor) {
```
oykuyilmaz, line 58: does this work only with ASCII characters? Should we expect it to work with others as well?

oykuyilmaz (approved):

PR #5149, opened 2023-05-02
Inline autocomplete tooltip UX changes

azmkercso:
*Issue #, if available:* N/A

*Description of changes:*
Changed UX of the inline tooltip for inline autocomplete.
Changes:
- Icon buttons (for previous and next completion)
- Displaying current completion's index and total completion index
- Nicer UX for key binding display
- Support dark mode
- Display tooltip on active line hover or always (with toggle and delayed display/hide)
- Separated tooltip and command code

It can be tried out in the kitchen sink demo, by using the Alt-C|Option-C keybinding to trigger its appearance together with the inline autocompletion. By default, the kitchen sink demo initialises the tooltip in hover mode, so it will be only displayed if the mouse is over the active line. This setting can be changed in the toolbar's overflow menu.

**Updated revision**:
- Handles editor session change
- Only render tooltip when the cursor is in the editor view and the tooltip can fit above the line properly
- Improved tooltip repositioning
- Keep editor focused after clicking on the tooltip


src/tooltip.js
```diff
@@ -76,7 +76,7 @@ class Tooltip {
         if (x != null && y != null)
             this.setPosition(x, y);
         if (!this.isOpen) {
-            this.getElement().style.display = "block";
+            this.getElement().style.display = "";
```
whazor, line 79: is this required? this impacts all tooltips right?

src/test/all_browser.js
```diff
@@ -25,9 +25,8 @@ var testNames = [
     "ace/editor_navigation_test",
     "ace/editor_text_edit_test",
     "ace/editor_commands_test",
+    "ace/ext/command_bar_test",
     "ace/ext/hardwrap_test",
-    "ace/ext/inline_autocomplete_test",
```
oykuyilmaz, line 29: I see we still have "inline_autocomplete_test", is this a mistake removing it from all browser tests?

skazantsev:
Minor bugs:
- When the tooltip is shown on the last line it doesn’t disappear if you leave the tooltip by moving the cursor down.
- When the tooltip is shown on the line 1-2 it may be shown off the screen. We might want to show the tooltip below the line.

src/tooltip.js
azmkercso: I reverted it. My original thought was to let it be defined by the CSS class, but this feels safer.

src/test/all_browser.js
azmkercso: I re-added it, along with some fixes to the tests there so they pass both in mockDOM and actual DOM.

azmkercso:
@skazantsev The things you mentioned are fixed in the last revision.

akoreman (approved):
lgtm, one concern I have with the inline autocomplete is that the suggestions are not communicated to screen reader users afaict, but that's outside the scope of this PR.

Issue #2164, opened 2014-09-26
Accessibility for screen readers


Menelion:
As per [this discussion](https://groups.google.com/forum/#!msg/ace-discuss/O1bSliAz8sg/9clCdpfIsSsJ), Ace is totally inaccessible for screen reader users.  
this highly impacts all visually impaired (especially, blind) developers which are quite numerous. And, as online courses are the most accessible learning form for such users, please make an effort to improve the accessibility of your great editor.  
If I could help somehow (testing, patching, ...), I'm open to collaboration.  
Thanks!

pamelafox:
I just got a bug report from a teacher of a blind student going through our curriculum on Khan Academy. This is what the problem was specifically:
"For some reason, his screen reading software will NOT read the code he inputs when he is reviewing it with his arrow keys. The software DOES read his code as he inputs it, but will not read in the review process with arrow keys."
Has any investigation gone into better screen reader support since this issue was opened?

Thanks!

gjtorikian:
I'd like to see if I can work on this, as it interests me for a variety of reasons. I'm curious to know what the expected technical behavior is. Like, is there a good source of information for what the DOM markup should look like so that a screen reader could process the information? Is an `aria-label` enough? I really don't know.

As @kevinb7 mentions elsewhere, it's probably possible to coat the current lines on screen with a `DIV` containing the underling information. IIRC, Ace calls these layers, and there's one for a bunch of different reasons:

![screen shot 2015-03-29 at 6 56 56 pm](https://cloud.githubusercontent.com/assets/64050/6889164/6cef2d42-d645-11e4-89ee-7cda981e3279.png)

Whether or not that's a good idea depends again on what the actual markup needs to be, though.

Oire:
Thanks @gjtorikian. Actually, the question is more difficult. I don't believe that `aria-label` will be enough because the input field contents is not read neither by arrow keys, nor when you, say, erase text with BackSpace. I wonder why that happens at all since if you use a plain `textarea` this doesn't happen. If we learn why this is so, we will be able to determine what (additional) markup is needed.

nightwing:
I think input field is not read because textarea in Ace is used only for capturing input and usually remains empty. 
Is there a way to trigger reading with javascript, e.g by setting aria-label on textarea or setting some text into textarea?
Ideally there would be a javascript api similar to chromevox so that we could do something like https://github.com/ajaxorg/ace/blob/master/lib/ace/ext/chromevox.js

[...]

PR #5160, opened 2023-05-09
Hide Ace content from assistive technologies to improve a11y

akoreman:
Issue #2164

Some screen reader/browser configurations read the entire visible part of the document when entering Ace. This aims to improve the screen reader experience by hiding the `ace_content` div from assistive technologies when a11y mode is enabled.


src/keyboard/textinput.js
```diff
@@ -65,14 +65,16 @@ var TextInput = function(parentNode, host) {
         }     
     };
     this.setAriaLabel = function() {
-        var row;
-        if (!host.session)
-            row = 0;
-        else
-            row = host.session.selection.cursor.row;
+        if(host.session && host.renderer.enableKeyboardAccessibility)
+        {
```
ibabanov-aws, line 69: Doesn't look in line with the code style

akoreman: ah whoops, you're right. Fixed in new commit

PR #5037, opened 2023-01-25
Implement highlight mode for PL/SQL (Oracle) dialect

rozhnev:
Description of changes:
Implemented new mode for highlight Oracle SQL dialect (PL/SQL)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5169, opened 2023-05-12
feat: add odin mode

laytan:
*Issue #, if available:* N/A

*Description of changes:* Added a basic mode for Odin


akoreman:
Thanks for your contribution, could you also add an example doc for the mode to the `docs/kitchen-sink/docs` folder and add it to the modelist in `src/ext/modelist.js`. For reference on what to add where, see PR #5037.

laytan:
@akoreman I have added the things you asked :)

akoreman:
Thanks :) I've been comparing your mode with the syntax highlighting found in the [Odin docs](https://odin-lang.org/docs/overview/) examples and it seems to match that pretty well.

One small quirk I noticed was highlighting for names which start with a capital (see picture, e.g. Vector3):

Screenshot 2023-05-14 192913

I don't know enough about Odin to tell what the desired behaviour is here but could you take a look at this?

Could you also add a test covering the indent/outdent behavior (for an example of such tests, see e.g. [here](https://github.com/ajaxorg/ace/blob/master/src/mode/ada_test.js)).

laytan:
@akoreman I've added the tests, let me know if they are ok like this.

Also removed the uppercase === constant rule because of the issue you were having and also that it can cause false-positives anyway.

akoreman:
Thanks for adding the tests, they look good to me. 

Correct me if I'm wrong, but I think you might have removed the wrong lines from the highlight file. AFAIK the lines you removed are for `#xyz_abc` (which are [Odin directives](https://odin-lang.org/docs/overview/#directives) if I read correctly). I think you meant to remove [these lines](https://github.com/ajaxorg/ace/pull/5169/files#diff-b400de2ec23b069012324dfbb7b5c558710e0c525f6fbc4ab56ded204cea2782R135-R138)?

InspiredGuy (approved):

valerio-123 (commented):
Mamma

valerio-123 (commented):
Mamma

PR #5204, opened 2023-06-13
feat(autocomplete): ellipsis is shown when suggestions is longer than popover width

oykuyilmaz:
When the completion suggestion is larger than the autocomplete popover width, we need to show elipsis(...) to convey this information.
This PR also makes right-alignment of meta default fixing #3785.

Screenshot 2023-06-13 at 16 43 01

Screenshot 2023-06-13 at 16 52 09

### DIFF
--- a/src/autocomplete/popup.js
+++ b/src/autocomplete/popup.js
@@ -443,10 +443,16 @@ dom.importCssString(`
 }
 .ace_autocomplete .ace_line {
     display: flex;
+    align-items: center;
 }
-.ace_autocomplete .ace_line .ace_ {
+.ace_autocomplete .ace_line > * {
+    min-width: 0;
+    flex: 0 0 auto;
     overflow: hidden;
     white-space: nowrap;
+}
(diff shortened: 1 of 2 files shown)


akoreman:
Do you know how screenreaders react to overflowed content shown by ellipsis? Does it read out the full content, read out the visual lines + something like 'more content available' or not read out the overflown content at all?

oykuyilmaz:
> 

Yes, they do read the whole content as some other text editor does. We can discuss what is preferred.

src/autocomplete/popup.js
```diff
@@ -438,13 +438,18 @@ dom.importCssString(`
     background: #25282c;
     color: #c1c1c1;
 }
-.ace_autocomplete_right .ace_text-layer  {
-    width: calc(100% - 8px);
+.ace_autocomplete .ace_text-layer  {
+    width: 100%;
```
oykuyilmaz, line 442: Current: 
with8

After this PR:
without8

PR #5205, opened 2023-06-13
fix: Improve ARIA labelling fold controls

akoreman:
*Description of changes:* For a11y compliance, buttons that show/hide content should have an `aria-expanded` state. This adds that state and changes the aria label accordingly.


oykuyilmaz (approved):

demo/i18n.html
```diff
@@ -72,9 +72,9 @@
         "Autocomplete suggestions": "Предложения автозаполнения",
                 
         "Cursor at row $0": "",
-        "Unfold rows $0 to $1": "",
+        "Code folding, rows $0 through $1": "",
+        "Code folding, row $0": "",
         "Unfold code": "",
-        "Fold at row $0": "",
```
InspiredGuy, line 77: if someone was using the old translation keys, what is going to happen? can we at least show a warning? can we support both by deprecating the old one?

akoreman: added a warning when a Ace tries to replace a string using the `nls` function but can't find a translation in the provided messages.

andredcoliveira: Does this mean this is a breaking change? If so, can we avoid it or is it intentional?

whazor (approved):
looks nice!

InspiredGuy (approved):

PR #5217, opened 2023-06-26
fix: bug in guttertooltip when `tooltipsFollowsMouse` set to false

akoreman:
*Issue #, if available:* NA

Fixes a bug where an error is thrown when hovering over the gutter with `tooltipsFollowsMouse == true` and scrolled. Additionally, adds `tooltipsFollowsMouse` to kitchen-sink to make it easier to test this option.


src/layer/lines.js
```diff
@@ -42,6 +42,15 @@ class Lines {
     get(index) {
         return this.cells[index];
     }
+
+    getCellAtRow(row) {
+        for (var i = 0; i < this.getLength(); i++){
```
whazor, line 47: this could be a binary search right? or maybe something faster

akoreman: changed the implementation of the search

src/mouse/mouse_event.js
```diff
@@ -48,6 +48,18 @@ class MouseEvent {
...
         return this.$pos;
     }
+
+    /**
+     * Get the relative position within the gutter.
+     * 
+     * @return {Number} 'row' within the gutter. 
+     */
+    getGutterRow() {
+        var documentRow = this.editor.renderer.screenToTextCoordinates(this.clientX, this.clientY).row;
```
ibabanov-aws, line 58: Might be better to use `this.getDocumentPosition().row` here

ibabanov-aws (approved):

src/mouse/mouse_event.js
akoreman: ah yeah, good catch, changed

src/layer/lines.js
akoreman: implementation is changed to not use any search anymore

ibabanov-aws (approved):

PR #5221, opened 2023-06-27
fix: set aria-live to textarea in a11y mode

akoreman:
*Issue #, if available:* NA

*Description of changes:* In some browsers, some screen readers don't pick up changes we make to the textarea resulting in the screen reader not reading lines when navigating up/down. Trying whether seting `aria-live` to the textarea to explicitly tell the screen reader to read the full line when cursor changes is an improvement and whether is has drawbacks. 


akoreman:
This does make screen readers which currently ignore text area content read it, but is has some drawbacks. Some screen readers now read content twice. Whether this is a better place to be than what we have currently is up to opinion. Something like #5225 might be a better way to go, but does require riskier changes.

akoreman:
Closing this PR since #5225 seems to be a better way to improve this.

PR #5225, opened 2023-06-28
feat: Allow more lines in hidden textarea to improve screen reader experience on Windows

akoreman:
*Issue #, if available:* #2164

*Description of changes:* Some Windows browsers and screen readers have trouble reading the content in the hidden textarea when navigating up/down.

This change allows more lines of text into the textarea to allow to improve the screen reader experience (without changing the default behavior).

This allows the cursor in the textarea to move lines up/down which makes the behavior in the hidden textarea closer to a 'normal' textarea. Monaco uses a similar approach and is generally considered to be ahead of Ace when it comes to screen reader compatibility.


akoreman:
This does seem to improve screen reader comparability, there are some problems with the current implementation when copying text though.

src/editor.js
```diff
@@ -2889,13 +2889,19 @@ config.defineOptions(Editor.prototype, "editor", {
...
+            // - Hide irrelevant elements from assistive technology.
+            // - On Windows, set more lines to the textarea.
             if (value){
                 this.renderer.enableKeyboardAccessibility = true;
                 this.renderer.keyboardFocusClassName = "ace_keyboard-focus";
 
                 this.textInput.getElement().setAttribute("tabindex", -1);
+                // VoiceOver on Mac OS works best with single line in the textarea, the screen readers on
+                // Windows work best with multiple lines in the textarea.
+                this.textInput.setNumberOfExtraLines(useragent.isWin ? 3 : 0);
```
whazor, line 2904: what about linux?

akoreman: Good question, I don't have quick access to a Linux desktop so I couldn't test this so I didn't want to change anything for Linux. The most used screen reader on Linux seems to be [orca](https://help.gnome.org/users/orca/stable/index.html.en).

src/keyboard/textinput.js
```diff
@@ -199,19 +224,49 @@ var TextInput = function(parentNode, host) {
...
+
+            for (var i = rowStart; i < row; i++) {
+                lines.push(host.session.getLine(i));
+            }
+
+            lines.push(host.session.getLine(row));
+
+            for (var i = row + 1; i <= rowEnd; i++) {
+                lines.push(host.session.getLine(i));
+            }
```
InspiredGuy, line 254: can it be simplified to this? seems like yes
```
for (var i = rowStart; i <= rowEnd; i++) {
    lines.push(host.session.getLine(i));
}
```

src/keyboard/textinput.js
```diff
@@ -63,21 +80,16 @@ var TextInput = function(parentNode, host) {
         if (options.role) {
             text.setAttribute("role", options.role);
         }     
-    };
-    this.setAriaLabel = function() {
-        if(host.session && host.renderer.enableKeyboardAccessibility) {
-            var row =  host.session.selection.cursor.row;
-
+        if (options.label) {
```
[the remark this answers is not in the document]
akoreman: yeah that's fair (I originally called it `setAriaLabel` but changed it to just `label` because the other one was too verbose, `setLabel` sounds like a good middleground)

src/keyboard/textinput.js
```diff
@@ -25,6 +25,7 @@ var TextInput = function(parentNode, host) {
     text.setAttribute("autocorrect", "off");
     text.setAttribute("autocapitalize", "off");
     text.setAttribute("spellcheck", false);
+    text.setAttribute("aria-multiline", true);
```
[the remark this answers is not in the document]
akoreman: [`aria-multiline`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiline) is just to indicate that pressing `Enter` inserts a new line instead of submitting a form, but `<textarea>` has that semantic by default so it's not really adding anything here, I'll just remove it.

src/keyboard/textinput.js
akoreman: yeah it can, in a previous revision I used each component of these 3 separately but now that I don't anymore we can combine them yeah

InspiredGuy (approved):

PR #5241, opened 2023-07-06
bug(autocomplete): fix a11y violations

oykuyilmaz:
Accessibility violations for autocompletion popover are fixed.
- `role=listbox` is moved to `$textLayer` of `ace_autocomplete` editor to make its direct children have `role=option`, otherwise there is `aria-required-children` violation (https://dequeuniversity.com/rules/axe/4.5/aria-required-children?application=axeAPI)
- `aria-hidden` set for `textarea` of autocomplete popover (there is `textarea` there since it is an editor instance).
- `aria-posinset` now starts from 1.

### DIFF
--- a/src/autocomplete/popup.js
+++ b/src/autocomplete/popup.js
@@ -52,6 +52,8 @@ class AcePopup {
         // Set aria attributes for the popup
         popup.renderer.container.setAttribute("role", "listbox");
         popup.renderer.container.setAttribute("aria-label", nls("Autocomplete suggestions"));
+        popup.renderer.container.setAttribute("aria-busy", "true");
+        popup.renderer.textarea.setAttribute("aria-hidden", "true");
 
         popup.setOption("displayIndentGuides", false);
         popup.setOption("dragDelay", 150);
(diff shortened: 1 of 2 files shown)


src/autocomplete/popup.js
```diff
@@ -52,6 +52,8 @@ class AcePopup {
         // Set aria attributes for the popup
         popup.renderer.container.setAttribute("role", "listbox");
         popup.renderer.container.setAttribute("aria-label", nls("Autocomplete suggestions"));
+        popup.renderer.container.setAttribute("aria-busy", "true");
```
InspiredGuy, line 55: you mentioned `aria-busy` is required due to required children rule described here https://dequeuniversity.com/rules/axe/4.5/aria-required-children?application=axeAPI
which element and with which role requires `aria-busy` to be set for it's child/parent/sibling? I don't see any explicit relations in [aria-busy documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy)

oykuyilmaz: Removed in commit 3

akoreman (approved):

PR #5258, opened 2023-07-17
Update lua.lua

first-leon:
Fix bug in example


akoreman:
Thanks for your contribution

PR #5269, opened 2023-07-26
Astro mode syntax highlighting

legendSabbir:
*Description of changes:*
# Syntax highlighting for Astro mode


akoreman:
Thanks for your contribution, could you also add an example doc for the mode to the demo/kitchen-sink/docs folder and add it to the modelist in src/ext/modelist.js. For reference on what to add where, see PR https://github.com/ajaxorg/ace/pull/5169.

legendSabbir:
@akoreman hey i've added a file to kitchen sink and also added in modelist . Anything else necessary ?

akoreman:
Could you:
- fix the issues thrown by the linter in the [built](https://github.com/ajaxorg/ace/actions/runs/5710937240/job/15475907734?pr=5269#step:8:75).
- remove the mode and highlight rules files from the `src` folder (they should only be in the `src/mode` folder).

legendSabbir:
@akoreman try now

legend-sabbir:
@akoreman fixed

akoreman:
thanks for fixing the build errors, mode looks good to me. Could you add some tests covering the mode (see for example [here](https://github.com/ajaxorg/ace/blob/master/src/mode/odin_test.js) for an example of tests for a mode)

legendSabbir:
I don't think i can do this . Isn't it optional 😕 ? if it's needed then Can you do it 😅

InspiredGuy:
The new Astro mode just inherits HTMLMode, so the tests for the mode itself are not really necessary. I'm merging this PR assuming one of the next mode-related PRs will regenerate tokens for highlight rule tests which will include Astro mode as well.

PR #5280, opened 2023-08-09
Fix completions replacement range calculation

mkslanc:
*Issue #, if available:*

*Description of changes:*

This PR addresses two primary concerns related to the completion replacement ranges:

1. Incorrect Replacement After Initial Letters:

**Issue:** Upon starting to type the beginning of the completion, the replacement range became unsynchronized by the second letter. This led to incorrect replacements, such as `<didiv` in place of `<div`.
**Solution:** Reverted the changes from [my prior PR](https://github.com/ajaxorg/ace/pull/5249), which resolved this issue but led to the second concern.

2. Deletion of Symbols Post-Cursor:

**Issue:** Given symbols following the cursor, like in between of "<>", when attempting to insert a tag between the symbols, the ">" would be erroneously deleted.
**Solution:** While the previously [reverted PR](https://github.com/ajaxorg/ace/pull/5279) addressed this issue, it also reintroduced the first issue.

To holistically address both concerns, i've introduced a new mechanism for calculating replacement ranges. Now, we preserve the initial cursor/base/prefix state prior to fetching completions. This stored state aids in accurately computing the necessary shift during the `insertMatch` process.

This revised method ensures both efficient replacement of initial letters and the preservation of subsequent symbols.

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -372,7 +372,6 @@ class Autocomplete {
         var completionOptions = { exactMatch: this.exactMatch };
         this.getCompletionProvider({
             prefix,
-            base: this.base,
             pos
         }).provideCompletions(this.editor, completionOptions, function(err, completions, finished) {
             var filtered = completions.filtered;
@@ -594,7 +593,7 @@ Autocomplete.startCommand = {
 class CompletionProvider {
 
     /**
-     * @param {{base: Anchor, pos: Position, prefix: string}} initialPosition
+     * @param {{pos: Position, prefix: string}} initialPosition
      */
     constructor(initialPosition) {
         this.initialPosition = initialPosition;
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -324,9 +324,9 @@ class Autocomplete {
         this.updateCompletions(false, options);
     }
 
-    getCompletionProvider() {
+    getCompletionProvider(initialPosition) {
         if (!this.completionProvider)
-            this.completionProvider = new CompletionProvider();
+            this.completionProvider = new CompletionProvider(initialPosition);
         return this.completionProvider;
     }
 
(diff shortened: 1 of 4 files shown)


src/autocomplete.js
```diff
@@ -611,20 +619,33 @@ class CompletionProvider {
             // TODO add support for options.deleteSuffix
```
whazor, line 619: Could you make the previous if case return (including `editor.endOperation();`)?

src/autocomplete_test.js
```diff
@@ -110,28 +117,93 @@ module.exports = {
...
+            afterRenderCheck(popup, function () {
                 editor.onCommandKey(null, 0, 13);
                 assert.equal(editor.getValue(), "goodwill-here");
                 editor.destroy();
                 editor.container.remove();
                 done();
             });
         });
+    },
+    "test: correct completion replacement range when filterText has more than one letter": function (done) {
```
whazor, line 134: where is `filterText` defined?

src/autocomplete.js
```diff
@@ -611,20 +619,33 @@ class CompletionProvider {
...
             if (!this.completions)
                 return false;
-            if (this.completions.filterText) {
+            
+            var replaceBefore = this.completions.filterText.length;
```
whazor, line 623: what does filterText do?

src/autocomplete.js
```diff
@@ -611,20 +619,33 @@ class CompletionProvider {
...
+            var replaceAfter = 0;
+            if (data.range && data.range.start.row === data.range.end.row) {
+                replaceBefore -= this.initialPosition.prefix.length;
```
whazor, line 626: Are we sure this doesn't go out of bounds? The logic is very complicated at the moment, and I don't see any guarantees.

src/autocomplete.js
```diff
@@ -611,20 +619,33 @@ class CompletionProvider {
...
+                    range.start.column -= replaceBefore;
+                    range.end.column += replaceAfter;
                     editor.session.remove(range);
                 }
             }
-            if (data.snippet)
-                snippetManager.insertSnippet(editor, data.snippet, {range: data.range});
+          
+            if (data.snippet) {
+                snippetManager.insertSnippet(editor, data.snippet, {range: undefined});
```
marinsokol5, line 647: Can you help me understand what is happening here, like why is it better if the range is undefined all of a sudden?

mkslanc: This shouldn't be there, I forgot to cleanup this after testing 👍

src/autocomplete_test.js
mkslanc: It's not defined, bad test naming, corrected it to reflect what actually is happening

src/autocomplete.js
mkslanc: Sorry, I don't understand your request. Do you mean 
```javascript
if (!this.completions) {
   editor.endOperation();
   return false;
}
```
to end already started operation in case there is no completions?

marinsokol5 (approved):

PR #5286, opened 2023-08-21
fix: autocompletion threshold doesn't apply to trigger characters

yaroslav-tykhonchuk:
*Issue #, if available:*

*Description of changes: When [Live autocompletion threshold](https://github.com/ajaxorg/ace/pull/5177/files) was introduced, it changed the behavior of [special trigger characters](https://github.com/ajaxorg/ace/pull/5147). 
Special trigger characters are an array of simple chars that should open the autocompletion popup immediately. And LiveAutocompletionThreshold applies to simple prefix, not trigger character.


uranium62:
This is exactly what I need

PR #5306, opened 2023-09-08
Lightweight tokenizer extension

marinsokol5:
Adding an Ace extension (`src/ext/simple_tokenizer.js`) exporting the smallest possible tokenizer which still uses Ace highlighting rules:
* Per my calculation it's below 10KB with the JSON rules. 
* This extension could be used by consumers to either wrap this into some kind of Syntax Highlighting component or just for parsing file use case, where a consumer is just interested in finding a particular token (like  `ace_variable`). It would be nice to extend on the docs for possible tokens though, the closest thing we have now is https://github.com/ajaxorg/ace/wiki/Creating-or-Extending-an-Edit-Mode#common-tokens AFAIK, but everyone who writes their own theme needs to know these.
* Some code did get shuffled around to accommodate that particular code path to make it that light, but there are no conceptual changes to those files.
     *  Core of `src/tokenizer.js` got moved to `src/tokenizer_internal.js` to avoid dependency on `src/config`.
     * `deepCopy()` got moved from `src/lib/lang.js` to `src/lib/deep_copy.js`
     * Text token logic got moved from `src/layer/text.js` to `src/layer/text_util.js` 
     * 


src/tokenizer_internal.js
akoreman, line 1: What's the reason for renaming this file?

marinsokol5: the smart logic got moved from tokenizer to tokenizer_internal so that tokenizer_internal doesn't depend on `config`. The only thing we actually use there is that `reportError` but still.

.eslintrc
```diff
@@ -2,6 +2,9 @@
     "parserOptions": {
         "ecmaVersion": 2015
     },
+    "env": {
+        "es6": true
```
marinsokol5, line 6: We can also just add Set to globals, but I don't see why we wouldn't have it here, since we also allow ES6 syntax with parserOptions.

src/tokenizer_internal.js
marinsokol5: I renamed the class to `TokenizerInternal` if that makes it more clear to not be used and that `Tokenizer` should be used. 
But yeah this reportError is basically monkey patching and is not really part of its interface, so no one actually using Tokenizer from outside of internal Ace modules will expect it to be there.

InspiredGuy: Renaming makes sense, but does not fix the issue. When the `tokenize` function added in this PR is called, the code paths still can lead to lines with `this.reportError`, which will result in `reportError is not a function` errors thrown. So even if we hint that `TokenizerInternal` is not meant to be used standalone and it won't be used, implementation of `SimpleTokenizer` still uses it - and may produce errors. I suggest to check if `this.reportError` exists in `TokenizerInternal` implementation before calling it everywhere the call occurs so we don't have to invest too much into refactoring of this logic now. This way this class will be available for standalone use as well.
```js
this.reportError && this.reportError(...);
```

marinsokol5: Aha, now I get it, yeah makes sense, I actually searched for who uses reportError outside of this class, but not internally... Jeesh.

src/ext/syntax-highlighter.js
```diff
@@ -0,0 +1,79 @@
...
+     */
+    constructor(content, tokenizer) {
+        this._lines = content.split(/\r\n|\r|\n/);
+        this._states = [];
+        this._tokenizer = tokenizer;
+    }   
+
+    /**
+     * @param {number} row 
+     * @returns {import("../../ace").Ace.Token[]}
```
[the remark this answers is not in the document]
marinsokol5: Yeah so for me in VSC it's needed yeah. I can't use Token or Ace.Token on their own, it just becomes `any`

InspiredGuy (approved):

PR #5321, opened 2023-09-21
Try rendering hover tooltip above mouse

whazor:
*Description of changes:*
Currently, Ace renders the hover tooltip below the mouse. But in the most common scenarios, where the user is typing code and using auto-complete, the hover tooltip tends to overlap the auto-complete.

screenshot of tooltip overlapping the autocomplete

By moving the hover tooltip above the mouse there is much less chance of overlapping with autocomplete, as the autocomplete renders below and the hover tooltip will go above. 

It would be possible to have a more advanced collision detection and try to render the hover tooltip in any position that is not blocking the autocomplete. But that could be a future improvement and this PR as is would already improve most use cases.

This PR also allows the tooltip to render outside the editor.

### DIFF
--- a/src/tooltip.js
+++ b/src/tooltip.js
@@ -282,28 +282,16 @@ class HoverTooltip extends Tooltip {
         element.style.display = "block";
         
         var position = renderer.textToScreenCoordinates(range.start.row, range.start.column);
-        var cursorPos = editor.getCursorPosition();
         
         var labelHeight = element.clientHeight;
         var rect = renderer.scroller.getBoundingClientRect();
 
-        var isTopdown = true;
-        if (this.row > cursorPos.row) {
-            // don't obscure cursor
-            isTopdown = true; 
-        } else if (this.row < cursorPos.row) {
-            // don't obscure cursor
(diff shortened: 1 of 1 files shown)


src/tooltip.js
```diff
@@ -282,28 +282,16 @@ class HoverTooltip extends Tooltip {
...
-        var isTopdown = true;
-        if (this.row > cursorPos.row) {
-            // don't obscure cursor
-            isTopdown = true; 
-        } else if (this.row < cursorPos.row) {
-            // don't obscure cursor
-            isTopdown = false;
-        }
-        
+        var isAbove = true;
```
marinsokol5, line 289: Can we use `const` and `let` in the newly added code? Or do we still want to cater for some really old browsers for Ace? Cause I think are eslint rules are set to ES6

nightwing (approved):

akoreman:
@whazor Do you have a screenshot of what the 'default' will look like after this change? (i.e. when you have a hover tooltip above a completion window)

akoreman (approved):

PR #5322, opened 2023-09-22
feat: Keep focus on same item in completion popup when slow completer delivers results.

akoreman:
*Issue #, if available:* NA

*Description of changes:* Currently, when there are new async completion results delivered by a completer, the active row of the popup is set back to 0. This is jarring when there are completers which are sufficiently slow that the user has already started to interact with the other, faster, completion results.

This adds a timer such that when completions are delivered 500ms after opening the popup, the item in focus remains in focus after the new results are added to the popup.

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -75,8 +75,8 @@ class Autocomplete {
 
         this.tooltipTimer = lang.delayedCall(this.updateDocTooltip.bind(this), 50);
 
-        this.keepSelectedRowTimer = lang.delayedCall(function() {
-            this.keepSelectedRow = true;
+        this.keepPreviouslySelectedRowTimer = lang.delayedCall(function() {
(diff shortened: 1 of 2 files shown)


src/autocomplete.js
```diff
@@ -62,6 +62,7 @@ class Autocomplete {
         this.keyboardHandler.bindKeys(this.commands);
         this.parentNode = null;
         this.setSelectOnHover = false;
+        this.stickySelectionDelay = 500;
```
marinsokol5, line 65: Can we add JSDoc comment somehow. for this option, since it's a bit hard to understand from just the name without prior context, but it's a pretty cool functionality.

akoreman: added a (hopefully) descriptive comment about this feature

marinsokol5 (approved):

PR #5327, opened 2023-09-25
Inline preview/ghost text screen reader a11y

akoreman:
*Issue #, if available:* NA

*Description of changes:* Currently, there is no way for screen reader users to learn the contents of the ghost text inline preview. This adds an `aria-describedby` to the completion item with inline preview which points to a screen reader only div displaying the contents of the ghost test. This allows the screen reader to read the content of the inline preview using the `aria-describedby` UX of their screen reader.

https://github.com/ajaxorg/ace/assets/34573235/d95f53dc-0aad-448e-84c1-3c4c2ec279da

### DIFF
--- a/src/autocomplete/inline_screenreader.js
+++ b/src/autocomplete/inline_screenreader.js
@@ -9,7 +9,7 @@ class AceInlineScreenReader {
 
     setScreenReaderContent(content) {
         // Path for when inline preview is used with 'normal' completion popup.
-        if (!this.popup && this.editor.completer && this.editor.completer.popup) {
+        if (!this.popup && this.editor.completer?.popup) {
             this.popup = this.editor.completer.popup;
 
             this.popup.renderer.on("afterRender", function() {
--- a/src/autocomplete/inline_screenreader.js
+++ b/src/autocomplete/inline_screenreader.js
@@ -8,7 +8,8 @@ class AceInlineScreenReader {
     }
 
     setScreenReaderContent(content) {
-        if (!this.popup) {
+        // Path for when inline preview is used with 'normal' completion popup.
+        if (!this.popup && this.editor.completer && this.editor.completer.popup) {
             this.popup = this.editor.completer.popup;
 
             this.popup.renderer.on("afterRender", function() {
@@ -24,6 +25,8 @@ class AceInlineScreenReader {
                 };
             }.bind(this));
         };
+
+        // TODO: Path for when special inline completion popup is used.
 
         // Remove all children of the div
         while (this.screenReaderDiv.firstChild) {
--- a/src/autocomplete/inline_screenreader.js
+++ b/src/autocomplete/inline_screenreader.js
@@ -51,11 +51,6 @@ class AceInlineScreenReader {
     }
 
     destroy() {
-        // Remove all children of the div
-        while (this.screenReaderDiv.firstChild) {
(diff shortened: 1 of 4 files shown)


akoreman:
Switched the implementation from using `aria-details` to `aria-describedby` due to the [lacking support](https://a11ysupport.io/tech/aria/aria-details_attribute) for `aria-details`.

src/autocomplete/inline_screenreader.js
```diff
@@ -0,0 +1,82 @@
...
+            this.screenReaderDiv.removeChild(this.screenReaderDiv.firstChild);
+        }
+        this._lines = content.split(/\r\n|\r|\n/);
+        const codeElement = this.createCodeBlock();
+        this.screenReaderDiv.appendChild(codeElement);
+    }
+
+    destroy() {
+        // Remove all children of the div
+        while (this.screenReaderDiv.firstChild) {
```
whazor, line 55: Is this needed when you remove the div itself below?

akoreman: What happens if you remove an element which has child nodes? (I was expecting bad things to happen, which is why I removed the child nodes first here)

whazor: They will be removed as well.

src/autocomplete_test.js
```diff
@@ -778,6 +778,48 @@ module.exports = {
...
+        // Popup should be open, with inline text renderered.
+        assert.equal(editor.completer.popup.isOpen, true);  
+        assert.equal(completer.popup.getRow(), 0);
+        assert.strictEqual(inline.isOpen(), true);
+        assert.strictEqual(editor.renderer.$ghostText.text, "function\nthat does something\ncool");
+
+        editor.completer.popup.renderer.$loop._flush();
+        var popupTextLayer = completer.popup.renderer.$textLayer;
+
+        // aria-details of selected popup item should have aria-details set to the offscreen inline screen reader div.
```
InspiredGuy, line 814: nit: it's not `aria-details` anymore, it's `aria-describedby`

src/autocomplete/inline_screenreader.js
```diff
@@ -0,0 +1,77 @@
...
+                    let idString = "doc-tooltip ";
+                    for (let lineIndex = 0; lineIndex < this._lines.length; lineIndex++) {
+                        idString += `ace-inline-screenreader-line-${lineIndex} `;
+                    }
+                    selected.setAttribute("aria-describedby", idString);      
+                }
+            }.bind(this));
+        }
+
+        // TODO: Path for when special inline completion popup is used.
```
InspiredGuy, line 42: is there a `special` inline completion popup?

src/autocomplete_test.js
akoreman: good catch, let me fix this

src/autocomplete/inline_screenreader.js
InspiredGuy: do we need to handle this case as part of this PR?

src/autocomplete_test.js
akoreman: fixed this

src/autocomplete/inline_screenreader.js
akoreman: Added GH issue to todo note in code.

PR #5342, opened 2023-10-04
feat: add nasal language

BobDotCom:
*Description of changes:*
Adds syntax highlighting for the Nasal scripting language.

Nasal (Not Another Scripting LAnguage) is FlightGear's built-in scripting language.
https://wiki.flightgear.org/Nasal
[http://plausible.org/nasal](https://web.archive.org/web/20150415110920/http://www.plausible.org:80/nasal/)
https://github.com/BobDotCom/Nasal.tmbundle


InspiredGuy:
Thank you for the contribution! The build has failed due to one of the regular expressions, could you please take a look? It would also be great if you could add a test for this newly added mode, there are examples of tests in https://github.com/ajaxorg/ace/tree/master/src/mode like https://github.com/ajaxorg/ace/blob/master/src/mode/javascript_test.js

BobDotCom:
> Thank you for the contribution! The build has failed due to one of the regular expressions, could you please take a look? It would also be great if you could add a test for this newly added mode, there are examples of tests in https://github.com/ajaxorg/ace/tree/master/src/mode like https://github.com/ajaxorg/ace/blob/master/src/mode/javascript_test.js

@InspiredGuy The regex should be fixed now. Do the tests need to be added now or can I add them in a later PR?

InspiredGuy:
@BobDotCom the build now fails due to linter errors, please take a look. We would prefer tests to be added to this PR, but your mode seems to just inherit the TextMode and not add any new behaviours so it should be fine as is. I would only ask to generate tokens for highlight rules test - see https://github.com/ajaxorg/ace/blob/master/src/mode/_test/Readme.md.

BobDotCom:
@InspiredGuy I fixed those linter errors and generated the tokens, does that look good?

InspiredGuy:
Looks good @BobDotCom , thank you

Issue #5338, opened 2023-10-02
Ace does not handle modern java features


bowbahdoe:
### Describe the bug

<img width="418" alt="image" src="https://github.com/ajaxorg/ace/assets/5004262/a6d46afb-beef-4653-9301-db9fde0b673a">

vs.

<img width="730" alt="image" src="https://github.com/ajaxorg/ace/assets/5004262/68df0965-8d02-4130-9d72-2ee30cf4b5ab">

### Expected Behavior

I'd expect things like `record` and multi-line string literals to be highlighted

### Current Behavior

`record` is not hightlighted and multi-line string literals are not recognized. Other things like patterns in switch statements or string templates might have similar issues

### Reproduction Steps

Make an editor in java mode, paste in newer code

### Ace Version / Browser / OS / Keyboard layout

1.2.8

andredcoliveira:
This is likely a matter of updating https://github.com/ajaxorg/ace/blob/master/src/mode/java_highlight_rules.js

bowbahdoe:
@andredcoliveira Not really... adding the `record` keyword would be about the cap of my ability to confidently do.

bowbahdoe:
I made a PR for that easy stuff

Issue #2742, opened 2015-11-01
Java syntax highlighting is not satisfying


vinothwindows47:
This is sublime java syntax highlighting mode function name is highlighting ....

![java-mode-syntax-highlighting-sublime](https://cloud.githubusercontent.com/assets/3929019/10867695/31a4c8d6-8096-11e5-8803-3d4705cd3ad7.png)

PR #5343, opened 2023-10-06
Update java highlight mode to support 21 version features

mkslanc:
*Issue #, if available:* #5338 #2742

*Description of changes:*


src/mode/java_highlight_rules.js
```diff
@@ -145,6 +121,143 @@ var JavaHighlightRules = function() {
             }, {
                 defaultToken : "comment"
             }
+        ],
+        "strings": [
+            {
+                token: "string",
+                regex: /(?:STR|FMT)\."/,
```
bowbahdoe, line 128: This isn't correct. Technically the prefix can be any expression so long as it returns a TemplateProcessor

so

```java
try (var conn = ds.getConnection()) {
   var stmt = conn."""
   SELECT * FROM table WHERE x = \{x}
   """;
}
```

or 

```java
a.b().c().e()."""
   Sub \{here}
"""

mkslanc: In that case, new approach should work. The only problem I see - is usage of outstanding ."""text""" without adding any processor, but for solving this case mode should recognize whole contexts and this would be a little bit too much.

InspiredGuy (approved):

bowbahdoe:
I see i was requested to review it - give me a minute to figure out how to test it locally...

InspiredGuy:
Hey @bowbahdoe thank you for the review! Would you consider the issue you highlighted above a deal breaker to merge this PR? I'm not a Java user myself so it's hard for me to evaluate the severity of the issue you pointed out above.

@mkslanc please take a look at the issue above, maybe fixing it would be faster than discussing it :)

mkslanc:
@bowbahdoe @InspiredGuy This change should make it :)

bowbahdoe:
@mkslanc I will take a look tonight

bowbahdoe (approved):
I think this looks good. Once there is a release, i'll update https://java-playground.com and if we find any more issues i'll report them

InspiredGuy:
Thank you @bowbahdoe , your help is appreciated!

Issue #2653, opened 2015-08-28
Hitting CTRL-F twice triggers Replace


snobu:
https://github.com/ajaxorg/ace/blob/master/lib/ace/ext/searchbox.js#L150

![image](https://cloud.githubusercontent.com/assets/6472374/9557995/73d1073a-4d95-11e5-83fe-e46077fc4215.png)

Is there any chance we can go back to just `CTRL-F` for search?
Hitting `CTRL-F` twice goes into Replace mode `(CTRL-H)` which ends up with accidental replacements pretty much every time.

Am i the only one who never used a text editor which had this `CTRL-F` behavior?

snobu:
Hey you fixed this. You sir are awesome!

PR #5346, opened 2023-10-07
Add "enableDoubleFindToReplace" option

nunziotocci:
*Issue #, if available:*
Brings back the behavior that #2653 complained about behind an option.

*Description of changes:*
Add "enableDoubleFindToReplace" option. This allows you to press Ctrl-F while the find window is open to toggle the replace option. I'm used to it, and so are the people at my company.


nunziotocci:
I have added it to `ace.d.ts` via `ace-internal.d.ts` and `npm run update-types`. I also added it to the Kitchen Sink.

I don't see any tests for `ext/searchbox`. How might I go about adding one?

PR #5353, opened 2023-10-12
Don't move popup with inline preview if setSelectOnHover true

akoreman:
*Issue #, if available:* NA

*Description of changes:* When `setSelectOnHover` is set to `True`, we select the item in the popup tooltip that's being hovered on. This can cause the height of the inline preview to change, resulting in moving the popup, after which the mouse can be hovering over a new item and selecting that. Leading to a cycle of moving popup:

https://github.com/ajaxorg/ace/assets/34573235/72e521e4-fcc1-403b-bbe2-1fc1845a533b

This aims to improve the situation by only moving the popup if it's not being moused over, and moving it to the desired position on `mouseout`:

https://github.com/ajaxorg/ace/assets/34573235/d9e1cb30-18c6-40a1-b3e2-46af59b7f247

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -319,8 +319,10 @@ class Autocomplete {
     }
 
     mouseOutListener(e) {
-        this.$updatePopupPosition();
-        this.tooltipTimer.call(null, null);
+        // Check whether the popup is still open after the mouseout event,
+        // if so, attempt to move it to its desired position.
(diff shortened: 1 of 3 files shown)


whazor (approved):
I tested the PR and it works nicely now!

Code looks good.

InspiredGuy (approved):

PR #5365, opened 2023-10-23
Astro highlighting mode fix

legendSabbir:
- I've fixed this issues long ago . I didn't thought it's gonna get merged so i didn't uploaded it


src/mode/astro_highlight_rules.js
```diff
@@ -2,82 +2,104 @@
...
     },
-    nextState: "start"
-  }, {
-    regex: /{/,
-    token: "paren.lparen",
-    push: "inline-js-start"
-  }]);
+    {
+      regex: /{/,
+      token: "paren.rparen",
```
akoreman, line 80: The regex here looks like this should be a left parenthesis?

paidPlugins: opps . the token should be  paren.lparen

PR #5366, opened 2023-10-23
fix: doc tooltip not filtering properly

akoreman:
*Issue #, if available:* NA

*Description of changes:* Fixes bug where the completion doc tooltip would not be filtered properly because it would get the data from the popup before it was updated:

https://github.com/ajaxorg/ace/assets/34573235/bdfa250e-1b27-44d4-8f7b-f802deda470e

after this change:

https://github.com/ajaxorg/ace/assets/34573235/fcbb1afe-42ac-4a0b-90dc-38a7f9b37adb

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -464,7 +464,7 @@ class Autocomplete {
 
     updateDocTooltip() {
         var popup = this.popup;
-        var all = popup.data;
+        var all = this.completions.filtered;
         var selected = all && (all[popup.getHoveredRow()] || all[popup.getRow()]);
         var doc = null;
         if (!selected || !this.editor || !this.popup.isOpen)
--- a/src/autocomplete_test.js
+++ b/src/autocomplete_test.js
@@ -297,6 +297,62 @@ module.exports = {
             }, 10);
         }
     },
+    "test: completers tooltip filtering": function (done) {
+        var editor = initEditor("");
+        var firstDoc = "First tooltip";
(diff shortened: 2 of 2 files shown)


src/autocomplete.js
```diff
@@ -464,7 +464,7 @@ class Autocomplete {
 
     updateDocTooltip() {
         var popup = this.popup;
-        var all = popup.data;
+        var all = this.completions.filtered;
         var selected = all && (all[popup.getHoveredRow()] || all[popup.getRow()]);
```
InspiredGuy, line 468: clearly, `popup.data` and `this.completions.filtered` don't match. but do `popup.getHoveredRow()` and `popup.getRow()` always correlate with `this.completions.filtered`?

InspiredGuy: also it would be great to write a test which would cover this scenario to prevent a regression

akoreman: Good question. When we call [`popup.setData`](https://github.com/ajaxorg/ace/blob/master/src/autocomplete/popup.js#L237), we transparently set whatever input we give it (`this.selections.filtered` [in this case](https://github.com/ajaxorg/ace/blob/master/src/autocomplete.js#L228)) to `popup.data`. So `popup.getRow`/`popup.getHoveredRow` should correlate with `this.completions.filtered`.

InspiredGuy (approved):

vanillajonathan:
Deploy on GitHub Pages!

Issue #5296, opened 2023-08-30
Yaml Mode: Missing Fold Widgets on list objects


robertpatrick:
### Describe the bug

When Ace is rendering YAML that contains list elements, the fold widget does not show unless the list is indented.  For example, the following YAML will display without a fold widget on the `properties` line.

```yaml
key: 32782-dtgeudkjes
properties:
- name: abc
  value: 123
- name: def
  value: 456
```
<img width="272" alt="Screenshot 2023-08-30 at 1 40 21 PM" src="https://github.com/ajaxorg/ace/assets/31662131/8124cbb2-3f49-460d-84b5-4db02fcfefb5">

Reformatting the YAML as shown causes the fold widget to appear on the `properties` line.

```yaml
key: 32782-dtgeudkjes
properties:
  - name: abc
    value: 123
  - name: def
    value: 456
```

<img width="275" alt="Screenshot 2023-08-30 at 1 41 17 PM" src="https://github.com/ajaxorg/ace/assets/31662131/41a7dd29-0b7c-4cae-88c6-db4165a771c4">

Unfortunately, a lot of YAML tools generate lists without the indention so this is pretty egregious bug. 

### Expected Behavior

I expect the fold widget to appear on the list object regardless of how the list is indented.

### Current Behavior

The list object has no fold widget if the list is not indented but does if the list is indented.

### Reproduction Steps

Drop the YAML snippets above in the editor and observe the fold widgets, as shown in the screen shots.

### Possible Solution

Indentation on subsequent lines cannot be the only thing that determines whether an object gets a fold widget or not.  For situations where there is no indentation, the code needs to check for `- ` (hyphen followed by a space) which would indicate that the line belongs to the list object defined above.

### Ace Version / Browser / OS / Keyboard layout

ace-builds@1.24.1, Chrome and Firefox, MacOS, US Keyboard layout

InspiredGuy:
Thank you for the report, seems like it should be fixed by creating a fold mode for Yaml extending the default one, which would check for `- ` as well.

PR #5370, opened 2023-10-25
fix: improve yaml folding 

akoreman:
*Issue #, if available:* #5296

*Description of changes:* WIP, to do:
- cleaning up a bit
- adding tests

### DIFF
--- a/src/mode/folding/yaml.js
+++ b/src/mode/folding/yaml.js
@@ -1,0 +1,90 @@
+"use strict";
+
+var oop = require("../../lib/oop");
+var BaseFoldMode = require("./fold_mode").FoldMode;
+var Range = require("../../range").Range;
+
+var FoldMode = exports.FoldMode = function() {};
+oop.inherits(FoldMode, BaseFoldMode);
+
+(function() {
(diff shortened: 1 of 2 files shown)


InspiredGuy (approved):

src/mode/folding/yaml.js
```diff
@@ -0,0 +1,108 @@
+"use strict";
+
+var oop = require("../../lib/oop");
+var CoffeeFoldMode = require("./coffee").FoldMode;
+var Range = require("../../range").Range;
+
+var FoldMode = exports.FoldMode = function() {};
+oop.inherits(FoldMode, CoffeeFoldMode);
```
oykuyilmaz, line 8: Since we inherit the CoffeeFoldMode, do we inherit its new `commentBlock`? Could it cause issues given CoffeeScript and YAML multi comments are different? 
e.g. 
```
###
This block 
comment
line
###
```
is not a valid block comment for YAML.

akoreman: It seems to work as desired:

no folding
Screenshot 2023-10-30 at 10 48 19

folding
Screenshot 2023-10-30 at 10 48 25

PR #5389, opened 2023-11-13
Fix empty search match before surrogate pair causing infinite loop

FIameCaster:
*Issue #, if available: NA*

*Description of changes:*
This PR fixes 2 issues relating to Ace's search.

The first fix is for an infinite loop. If the `u` flag is present and the regex returns an empty match before a surrogate pair, the `lastIndex` must be increased by 2 instead of 1.

The second fix relates to the `u` flag feature detection. Currently `$supportsUnicodeFlag` is set to `false` if the user types something that's an invalid regular expression. Once it's set to `false`, it can't be set to `true` again, effectively disabling the `u` flag even in browsers supporting it. This PR fixes these issues and removes the `supportsUnicodeFlag` method from `lang` since it's no longer needed.

### DIFF
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -167,6 +167,7 @@ class SearchBox {
     updateCounter() {
         var editor = this.editor;
         var regex = editor.$search.$options.re;
+        var supportsUnicodeFlag = editor.$search.$options.$supportsUnicodeFlag;
         var all = 0;
         var before = 0;
         if (regex) {
@@ -188,7 +189,8 @@ class SearchBox {
                 if (all > MAX_COUNT)
                     break;
                 if (!m[0]) {
-                    regex.lastIndex = last += 1;
+                    regex.lastIndex = last +=
+                        supportsUnicodeFlag && (value.charCodeAt(last) ^ 0xd800) < 1024 ? 2 : 1;
                     if (last >= value.length)
                         break;
                 }
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -167,7 +167,7 @@ class SearchBox {
     updateCounter() {
         var editor = this.editor;
         var regex = editor.$search.$options.re;
-        var supportsUnicodeFlag = editor.$search.$options.$supportsUnicodeFlag;
+        var supportsUnicodeFlag = regex.unicode;
         var all = 0;
         var before = 0;
         if (regex) {
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -189,8 +189,7 @@ class SearchBox {
                 if (all > MAX_COUNT)
                     break;
                 if (!m[0]) {
-                    regex.lastIndex = last +=
-                        supportsUnicodeFlag && (value.charCodeAt(last) ^ 0xd800) < 1024 ? 2 : 1;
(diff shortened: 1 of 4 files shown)


FIameCaster:
I also decided to remove the `$supportsUnicodeFlag` internal property entirely in favor of just using the `unicode` property on the regex for simplicity. In unsupporting browsers, it will return `undefined` which works perfectly for our use case. If `$supportsUnicodeFlag` should be kept, undo the last 2 commits.

src/ext/searchbox.js
```diff
@@ -188,7 +189,8 @@ class SearchBox {
                 if (all > MAX_COUNT)
                     break;
                 if (!m[0]) {
-                    regex.lastIndex = last += 1;
+                    regex.lastIndex = last +=
+                        supportsUnicodeFlag && (value.charCodeAt(last) ^ 0xd800) < 1024 ? 2 : 1;
```
InspiredGuy, line 193: I see this check repeated 3 times in this PR and it may look like magic to people unfamiliar with bitwise XOR. My understanding is it checks if the character falls into a particular range of unicode characters (range of 10 last bits starting with 0xd800) to determine if it's the "surrogate character", which is unusable on it's own. Let's extract it to a function with a meaningful name and leave a comment there explaining what the function does, this will be helpful in a long term.

InspiredGuy (commented):
Thank you for the contribution! Could you please also add tests for the fixed bugs to ensure we don't introduce a regression in the future?

src/ext/searchbox.js
FIameCaster: I did this to avoid using `codePointAt` which has worse browser support. But since only browsers that support the `u` flag will run that code, we can use it. With `codePointAt`, the code could be written like this:

```javascript
regex.lastIndex = last += supportsUnicodeFlag && value.codePointAt(last) > 0xffff ? 2 : 1;
```

Is this readable enough, or should it still be extracted to its own function?

InspiredGuy: I'd prefer it extracted anyway, to avoid unnecessary duplication - it's the same check in all the places. And it will be easier to cover that specific function with tests. I have no strong preference for implementation, so I'll leave it up to you - as long as it works and there are tests in place to prevent a regression :)

FIameCaster: Got it. Is `lib/lang` a good place to put it?

And lastly, is there a way to run a subset of the tests so I don't have to wait minutes every time I want to test new tests?

InspiredGuy: > Is lib/lang a good place to put it?

I think yes.

> And lastly, is there a way to run a subset of the tests so I don't have to wait minutes every time I want to test new tests?

Try running the tests in the browser by serving http://localhost:8888/src/test/tests.html and selecting a specific test suite there, that is what I used to do for this purpose. Let me know if it doesn't work out.

FIameCaster: Done. Both of the added tests have been verified to fail on the current version of Ace.

InspiredGuy (approved):

src/ext/searchbox.js
InspiredGuy: Thank you, the PR was merged 🙂

PR #5400, opened 2023-11-22
fix: try to scroll inline preview into view

akoreman:
*Issue #, if available:* NA

*Description of changes:* Currently, when inline preview is used at the bottom of the file, it will be invisible. This changes it to scroll down enough to get the inline preview into view. If the inline preview is bigger than the view, we scroll down so that the cursor is at the top of the screen to get the most possible into view

https://github.com/ajaxorg/ace/assets/34573235/9071dfcc-4127-4806-abb3-697c0ce2dcd0

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -92,7 +92,6 @@ class Autocomplete {
         this.tooltipTimer = lang.delayedCall(this.updateDocTooltip.bind(this), 50);
         this.popupTimer = lang.delayedCall(this.$updatePopupPosition.bind(this), 50);
 
-
         this.stickySelectionTimer = lang.delayedCall(function() {
             this.stickySelection = true;
         }.bind(this), this.stickySelectionDelay);
@@ -145,6 +144,7 @@ class Autocomplete {
         }
         this.hideDocTooltip();
         this.stickySelectionTimer.cancel();
+        this.popupTimer.cancel();
         this.stickySelection = false;
     }
 
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -82,6 +82,8 @@ class Autocomplete {
         }.bind(this));
 
         this.tooltipTimer = lang.delayedCall(this.updateDocTooltip.bind(this), 50);
+        this.popupTimer = lang.delayedCall(this.$updatePopupPosition.bind(this), 50);
+
 
         this.stickySelectionTimer = lang.delayedCall(function() {
             this.stickySelection = true;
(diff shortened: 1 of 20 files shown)


src/virtual_renderer.js
```diff
@@ -1620,6 +1620,24 @@ class VirtualRenderer {
                 className: "ace_ghost_text"
             };
             this.session.widgetManager.addLineWidget(this.$ghostTextWidget);
+
+            // Check wether the line widget fits in the part of the screen currently in view
+            var pixelPosition = this.$cursorLayer.getPixelPosition(insertPosition, true);
+            var el = this.container;
+            var height = el.clientHeight;
```
andredcoliveira, line 1627: `clientHeight` might be rounded down. Maybe using `getBoundingClientRect` would be a little more robust. Though I'm not sure how prominent fractional heights are in ace, if at all.

From https://developer.mozilla.org/en-US/docs/Web/API/Element/clientHeight

src/virtual_renderer.js
```diff
@@ -1620,6 +1620,24 @@ class VirtualRenderer {
...
+            var fitsY = textLines.length * this.lineHeight < height - pixelPosition.top;
+
+            // If it fits, no action needed
+            if (fitsY) return;
+            
+            // If it can fully fit in the screen, scroll down until it fits on the screen
+            // if it cannot fully fit, scroll so that the cursor is at the top of the screen
+            // to fit as much as possible.
+            if (textLines.length * this.lineHeight < height) {
```
andredcoliveira, line 1636: **nitpick:** we multiply `textLines.length * this.lineHeight` twice, so we can store the value and reuse it across lines 1628 and 1636.

src/autocomplete.js
```diff
@@ -160,9 +162,14 @@ class Autocomplete {
                 this.tooltipTimer.call(null, null);
                 return;
             }
+
+            // Update the popup position after a short wait to account for potential scrolling
+            this.popupTimer.schedule();
+            this.tooltipTimer.schedule();
+        } else {
+            this.$updatePopupPosition();
```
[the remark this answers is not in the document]
akoreman: `this.tooltipTimer` updates the position of the doc tooltip, `this.popupTimer` updates the position of the popup itself. They are both `lang.delayedCall`s which can be scheduled to happen after a timeout to allow some time for other stuff to happen.

andredcoliveira (approved):
Looks good. Left a couple of small comments.

src/autocomplete.js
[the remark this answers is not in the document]
whazor: Why does it happen outside the `this.inlineRenderer && this.inlineEnabled` case? Is it to cancel if inline rendering gets disabled? Or is it a general improvement for the non-inline case? If so, we should update the GH description.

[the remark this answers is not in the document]
akoreman: tried to make this a bit more clear in a new commit

src/virtual_renderer.js
akoreman: changed this

src/virtual_renderer.js
akoreman: changed this

andredcoliveira (approved):

PR #5401, opened 2023-11-27
Popup anchor

akoreman:
*Issue #, if available:* NA

*Description of changes:* Currently, when ghost text extends beyond the border of the editor, the autocomplete popup can be rendered disconnected from the editor:

https://github.com/ajaxorg/ace/assets/34573235/8863711a-f212-4fd1-9229-337a3b01d2ca

This changes this so that it will be rendered below, but still connected to, the editor at the lowest:

https://github.com/ajaxorg/ace/assets/34573235/c9b80240-05f2-4213-9a68-3614226d89b1

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -224,8 +224,15 @@ class Autocomplete {
             }
         }
 
+        // posGhostText can be below the editor rendering the popup away from the editor.
+        // In this case, we want to render the popup such that the top aligns with the bottom of the editor.
+        var editorContainerBottom = editor.container.getBoundingClientRect().bottom - lineHeight;
+        var lowestPosition = editorContainerBottom < posGhostText.top ?
+            {top: editorContainerBottom, left: posGhostText.left} :
+            posGhostText;
+
         // Try to render below ghost text, then above ghost text, then over ghost text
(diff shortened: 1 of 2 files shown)


src/autocomplete.js
```diff
@@ -231,8 +231,15 @@ class Autocomplete {
             }
         }
 
+        // posGhostText can be below the editor rendering the popup away from the editor.
+        // In this case, we want to render the popup such that the top aligns with the bottom of the editor.
+        var editorContainerBottom = editor.container.getBoundingClientRect().bottom - lineHeight;
+        var lowestPosition = editorContainerBottom < posGhostText.top ?
+            {top: editorContainerBottom, left: posGhostText.left} :
```
whazor, line 238: > "Added line #L238 was not covered by tests"

But you have a test that tests these changes, in particular it tests whether "Popup should start one pixel below the bottom of the editor". 

Do you think it would be valuable to also test the other edge case in your test?

akoreman: you mean the edge case where the popup is within the editor and should therefore be rendered at the bottom of the ghost text?

akoreman: added an extra test to the PR

oykuyilmaz (approved):

Issue #5377, opened 2023-10-31
Autocomplete: Screen reader doesn't read suggestions on Safari


oykuyilmaz:
### Describe the bug

Triggering autocomplete and navigating completions is not mentioned in screen reader on Safari.

### Expected Behavior

Autocomplete popover is accessible to screen reader users.

### Reproduction Steps

1. Open https://ace.c9.io/kitchen-sink.html
2. Start VoiceOver
3. Ctrl + Space to trigger autocomplete

### Ace Version / Browser / OS / Screen Reader

1.31.1/Safari/MacOS/VoiceOver

oykuyilmaz:
Works as expected on Chrome.

marinsokol5:
Can confirm that Chrome and Firefox work correctly but Safari 16.6 doesn't.  
Unsure if this is a Safari problem but we should definitely investigate if there is something we can do for those users.

oykuyilmaz:
We use `aria-activedescendant` attribute for the current selection on autocomplete popover with `listbox` role. Screen reader support expectations are as following for this attribute:
https://a11ysupport.io/tech/aria/aria-activedescendant_attribute

PR #5403, opened 2023-11-28
fix: set completion popup role to 'menu' for safari

akoreman:
*Issue #, if available:* #5377

*Description of changes:* On Safari, `aria-activedescendant` is [not supported](https://a11ysupport.io/tech/aria/aria-activedescendant_attribute) for the `listbox` role but is supported for the `menu` role. This results in an unusable autocomplete for VoiceOver/Safari users. This changes the role to `menu` for Safari to give some (but not perfect) support for this browser/screenreader combo:

https://github.com/ajaxorg/ace/assets/34573235/87065a11-b670-4e21-8f99-9397e20f7a67

Sets `aria-roledescription` to keep the behaviour somewhat consistent between the two flows. Sets `aria-selected` for the selected item, without it Safari/VoiceOver doesn't read the item.

### DIFF
--- a/src/autocomplete/popup.js
+++ b/src/autocomplete/popup.js
@@ -7,6 +7,7 @@ var event = require("../lib/event");
 var lang = require("../lib/lang");
 var dom = require("../lib/dom");
 var nls = require("../config").nls;
+var userAgent = require("./../lib/useragent");
 
 var getAriaId = function(index) {
     return `suggest-aria-id:${index}`;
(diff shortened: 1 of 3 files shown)


InspiredGuy:
The experience in other browsers is not affected by setting the `aria-selected` property, right?

akoreman:
@InspiredGuy not as far as I can tell, adding `aria-selected` makes the screenreader read `selected` on the selected item but Chrome and FF were already doing that.

oykuyilmaz (approved):

PR #5415, opened 2023-12-06
fix: switch aria-selected to aria-current for webkit

akoreman:
*Issue #, if available:* NA

*Description of changes:* follow-up to #5403, we set the role to `menuitem` for items in the completer popup on webkit browsers but `aria-selected` is not allowed for `menuitem` roles triggering customers who run automated a11y tests. This changes it to [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) which is allowed for this role.

### DIFF
--- a/src/autocomplete/popup.js
+++ b/src/autocomplete/popup.js
@@ -130,7 +130,10 @@ class AcePopup {
             if (selected !== t.selectedNode && t.selectedNode) {
                 dom.removeCssClass(t.selectedNode, "ace_selected");
                 el.removeAttribute("aria-activedescendant");
-                selected.removeAttribute("aria-selected");
+                if (userAgent.isSafari)
+                    selected.removeAttribute("aria-current");
+                else 
+                    selected.removeAttribute("aria-selected");
                 t.selectedNode.removeAttribute("id");
             }
             t.selectedNode = selected;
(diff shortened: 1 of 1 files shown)


src/autocomplete/popup.js
```diff
@@ -130,7 +130,10 @@ class AcePopup {
             if (selected !== t.selectedNode && t.selectedNode) {
                 dom.removeCssClass(t.selectedNode, "ace_selected");
                 el.removeAttribute("aria-activedescendant");
-                selected.removeAttribute("aria-selected");
+                if (userAgent.isSafari)
+                    selected.removeAttribute("aria-current");
+                else 
+                    selected.removeAttribute("aria-selected");
```
andredcoliveira, line 136: **nitpick:** seems like a great use-case for a ternary operator

```suggestion
                selected.removeAttribute(userAgent.isSafari ? "aria-current" : "aria-selected");
```

src/autocomplete/popup.js
```diff
@@ -146,7 +149,10 @@ class AcePopup {
                 selected.setAttribute("aria-setsize", popup.data.length);
                 selected.setAttribute("aria-posinset", row+1);
                 selected.setAttribute("aria-describedby", "doc-tooltip");
-                selected.setAttribute("aria-selected", "true");
+                if (userAgent.isSafari)
+                    selected.setAttribute("aria-current", "true");
+                else 
+                    selected.setAttribute("aria-selected", "true");
```
andredcoliveira, line 155: **nitpick:** seems like a great use-case for a ternary operator

```suggestion
                selected.setAttribute(userAgent.isSafari ? "aria-current" : "aria-selected", "true");
```

andredcoliveira (approved):
Looks good to me

PR #5416, opened 2023-12-07
fix: switch aria-selected to aria-current for webkit

akoreman:
*Issue #, if available:* NA

*Description of changes:* follow-up to #5403, we set the role to `menuitem` for items in the completer popup on webkit browsers but `aria-selected` is not allowed for `menuitem` roles triggering customers who run automated a11y tests. This changes it to [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) which is allowed for this role.

https://github.com/ajaxorg/ace/assets/34573235/f44da281-dcf1-46eb-8aa4-9c6a9f563d30


src/autocomplete/popup.js
```diff
@@ -153,7 +153,7 @@ class AcePopup {
                 selected.setAttribute("aria-setsize", popup.data.length);
                 selected.setAttribute("aria-posinset", row + 1);
                 selected.setAttribute("aria-describedby", "doc-tooltip");
-                selected.setAttribute("aria-selected", "true");
+                selected.setAttribute(userAgent.isSafari ? "aria-current" : "aria-selected", "true");
```
marinsokol5, line 156: We might wanna do the `isSafari` check more above to group Safari related changes in one place.

marinsokol5 (approved):

src/autocomplete/popup.js
akoreman: @sokomari refactored the PR a bit to make it clearer, do you think this is an improvement?

PR #5435, opened 2023-12-21
Update Logtalk support to the latest language spec

pmoura:
*Description of changes:*

- Add syntax highlight support for the `threaded_cancel/1` and `uninstantiation_error/1` built-in methods
- Add escape sequences support for double-quoted terms


src/mode/logtalk_highlight_rules.js
```diff
@@ -86,7 +74,7 @@ var LogtalkHighlightRules = function() {
          { token: 'support.function.control.logtalk',
            regex: '\\b(?:true|fa(?:il|lse)|repeat|(?:instantiation|system)_error)\\b(?![-!(^~])' },
          { token: 'support.function.control.logtalk',
-           regex: '\\b((?:type|domain|existence|permission|representation|evaluation|resource|syntax)_error)(?=[(])' },
+           regex: '\\b((?:uninstantiation|type|domain|existence|permission|representation|evaluation|resource|syntax)_error)(?=[(])' },
```
oykuyilmaz, line 77: do we need here `instantiation_error` as well? [Ref](https://logtalk.org/manuals/refman/methods/error_handling_methods.html).

pmoura: No. `instantiation_error/0` is above that line as it doesn't have any arguments unlike e.g. `uninstantiation_error/1` that have one argument (and thus `uninstantiation_error` must be followed by a `(` character; no spaces between allowed).

src/mode/logtalk_highlight_rules.js
```diff
@@ -170,7 +158,8 @@ var LogtalkHighlightRules = function() {
          { token: 'punctuation.definition.string.begin.logtalk',
            regex: '"',
            push: 
-            [ { token: 'constant.character.escape.logtalk', regex: '\\\\.' },
+            [ { token: 'constant.character.escape.logtalk',
+                regex: '\\\\([\\\\abfnrtv"\']|(x[a-fA-F0-9]+|[0-7]+)\\\\)' },
```
oykuyilmaz, line 162: should we consider other escape sequences like `\UXXXXXXXX` or `\uXXXX`?

pmoura: Those escape sequences are implemented only by some of the backend Prolog compilers supported by Logtalk, notably only for those that support Unicode. They are slowly becoming a de facto standard feature. But they are also rarely used.

oykuyilmaz (approved):

PR #5439, opened 2023-12-29
fix: improve prefixing for inline preview

akoreman:
*Issue #, if available:* NA

*Description of changes:* Use the string of text between the start of the autocomplete prefix and the current cursor position, regardless of the characters in between, to determine whether inline preview should be rendered. This allows inline preview to stay visible when typing ahead with whitespace.

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -197,7 +197,14 @@ class Autocomplete {
     }
 
     $updateGhostText(completion) {
-        var prefix = util.getCompletionPrefix(this.editor);
+        // Ghost text can include characters normally not part of the prefix (e.g. whitespace).
+        // When typing ahead with ghost text however, we want to simply prefix with respect to the
+        // base of the completion.
+        var row = this.base.row;
+        var column = this.base.column;
+        var cursorColumn = this.editor.getCursorPosition().column;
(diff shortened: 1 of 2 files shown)


src/autocomplete.js
```diff
@@ -197,7 +197,14 @@ class Autocomplete {
     }
 
     $updateGhostText(completion) {
-        var prefix = util.getCompletionPrefix(this.editor);
+        // Ghost text can include characters normally not part of the prefix (e.g. whitespace).
```
oykuyilmaz, line 200: do you know why `getCompletionPrefix` in utils doesn't consider whitespaces? Should we update that too?

src/autocomplete_test.js
```diff
@@ -1383,6 +1383,51 @@ module.exports = {
                 assert.strictEqual(inline.isOpen(), true);
                 assert.strictEqual(editor.renderer.$ghostText.text, "n\nthat does something\ncool");
     
+                done();
+            }, 100);
+        }, 100);
+    },
+    "test: should keep shwoing ghost text when typing ahead with whitespace": function(done) {
```
oykuyilmaz, line 1390: nit: `shwoing`

akoreman: 😢

oykuyilmaz (approved):

PR #5448, opened 2024-01-05
Add all ANSI Keywords to sql_highlight_rules.js

### LINKED ISSUE #5446: Enhance SQL Support
### Describe the feature

SQL has lots of keywords but only very few are already added. I think all ANSI Keywords should be there: https://docs.databricks.com/en/sql/language-manual/sql-ref-reserved-words.html#ansi-reserved-words 

Also some dialects have additional data types, like `string` in Spark.

### Use Case

Just some nicer syntax highlighting

### Proposed Solution

I could do a PR do add the keywords/datatypes

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### ACE version used

v1.32.3

aersam:
*Issue #, if available:*
Fixes #5446 

*Description of changes:*
Adds all ANSI Keywords and also adds string as datatype which is common in Spark and some other newer systems like DuckDB

### DIFF
--- a/src/mode/sql_highlight_rules.js
+++ b/src/mode/sql_highlight_rules.js
@@ -8,7 +8,13 @@ var SqlHighlightRules = function() {
     var keywords = (
         "select|insert|update|delete|from|where|and|or|group|by|order|limit|offset|having|as|case|" +
         "when|then|else|end|type|left|right|join|on|outer|desc|asc|union|create|table|primary|key|if|" +
-        "foreign|not|references|default|null|inner|cross|natural|database|drop|grant|distinct|is|in"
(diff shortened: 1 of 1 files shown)


src/mode/sql_highlight_rules.js
```diff
@@ -22,7 +28,7 @@ var SqlHighlightRules = function() {
 
     var dataTypes = (
         "int|numeric|decimal|date|varchar|char|bigint|float|double|bit|binary|text|set|timestamp|" +
-        "money|real|number|integer"
+        "money|real|number|integer|string"
```
whazor, line 31: is string part of the ANSI specifications? or only in duckdb?

aersam: Its not in the spec but very common

Issue #5447, opened 2024-01-04
JSX Mode: incorrectly highlights inner content


rubencarvalho:
### Describe the bug

JSX code is incorrectly highlighted (the element's inner content is being tokenised).

<img width="446" alt="Screenshot 2024-01-04 at 10 57 57" src="https://github.com/ajaxorg/ace/assets/29541876/bd0ad0f3-467b-472d-b8c5-1b4c71a7fa02">

### Expected Behavior

None of the words in the inner content of the elements should be highlighted.
GitHub's JSX syntax highlighting shows it best:
```jsx
import * as React from "react";
export default () => {
  return (
    <div variant="p">
      When you launch an instance, the instance class that
      you specify determines the hardware of the host
      computer used for your instance. Each instance class
      offers different compute, memory, and storage
      capabilities. Choose an instance class based on the
      requirements of the application or software that you
      plan to run on your instance.
    </div>
  );
}

```

### Reproduction Steps

1. Go to https://ace.c9.io/build/kitchen-sink.html
2. Set the mode to `JSX`
3. Delete the document's contents
4. Paste the following: 
```jsx
import * as React from "react";
export default () => {
  return (
    <div variant="p">
      When you launch an instance, the instance class that
      you specify determines the hardware of the host
      computer used for your instance. Each instance class
      offers different compute, memory, and storage
      capabilities. Choose an instance class based on the
      requirements of the application or software that you
      plan to run on your instance.
    </div>
  );
}
``` 

### Possible Solution

Fix the JSX highlighting rules.

### Ace Version / Browser / OS / Keyboard layout

Ace Version 1.32.3 (latest)

InspiredGuy:
Hey, thank you for the report. This doesn't look nice indeed. You could try to substitute `JSX` mode for `TSX` while we look into it.

InspiredGuy:
https://github.com/ajaxorg/ace/pull/5451 with a fix for this issue was merged and will be available in the next release.

PR #5451, opened 2024-01-08
fix: update JSX mode to reflect React JSX

InspiredGuy:
*Issue #, if available:*
https://github.com/ajaxorg/ace/issues/5447
*Description of changes:*
Existing JSX mode was introduced in 2012 for a project which seems inactive now (see PR https://github.com/ajaxorg/ace/pull/794). This PR:
- Updates JSX mode to use JS mode with `{jsx: true}` option to make it work for React JSX, which is pretty much what is associated with the word JSX nowadays. Also disables jshint for JSX mode since it does not work well with JSX.
- Makes `JSX` mode default for `*.jsx` files by removing this extension from being associated with JS mode.


whazor (approved):

rubencarvalho:
Thank you for the quick fix!

PR #5467, opened 2024-01-17
Add TypeScript types for modelist extension

nwalters512:
*Description of changes:* This PR introduces types for the modelist extension.

### DIFF
--- a/ace-extensions.d.ts
+++ b/ace-extensions.d.ts
@@ -12,7 +12,6 @@ declare module "ace-code/src/ext/modelist" {
         caption: string;
         mode: string;
         extensions: string;
-        extRe: RegExp;
         supportsFile(filename: string): boolean;
     }
     export function getModeForPath(path: string): Mode;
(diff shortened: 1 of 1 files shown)


ace-extensions.d.ts
```diff
@@ -5,3 +5,17 @@ declare module "ace-code/src/ext/simple_tokenizer" {
...
     export function tokenize(content: string, highlightRules: import("./ace").Ace.HighlightRules): TokenizeResult;
 }
+
+declare module "ace-code/src/ext/modelist" {
+    export type Mode = {
+        name: string;
+        caption: string;
+        mode: string;
+        extensions: string;
+        extRe: RegExp;
```
InspiredGuy, line 15: I think we could treat `extRe` as an internal property of the `Mode` (and maybe update implementation to have it as `_extRe`), since it's not used anywhere outside the `Mode` class and I think it shouldn't be - there is `supportsFile` method for it.

InspiredGuy: So I would suggest to treat it as internal and not expose it in a type.

nwalters512: Sounds good, done in 3bb0bc2.

PR #5510, opened 2024-03-15
Comments in JSON

Googleboom:
Issue #, if available:
JSON highlight rules allow comments '/**/' but the worker does not.

Description of changes:
The patch adds this feature to JSON worker.

Pull Request Checklist:

### DIFF
--- a/lib/ace/mode/json/json_parse.js
+++ b/lib/ace/mode/json/json_parse.js
@@ -189,10 +189,24 @@ define(function(require, exports, module) {
 
         white = function () {
 
-// Skip whitespace.
+// Skip whitespace and comments /**/.
 
-            while (ch && ch <= ' ') {
-                next();
+            while (true) {
+                while (ch && ch <= ' ') {
+                    next();
+                }
+
+                if (ch && ch === '/' && text.charAt(at) === '*') {
+                    next('/');
+                    next('*');
+                    while (ch && ch !== '*') {
+                        next();
+                    }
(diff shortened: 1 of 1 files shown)


andredcoliveira:
As far as I'm aware, comments are not part of the standard JSON specification: https://datatracker.ietf.org/doc/html/rfc8259.

For comments, either [JSON5](https://github.com/ajaxorg/ace/blob/master/src/mode/json5.js) can be used, or JSONC could be added to the list of Ace modes

Googleboom:
> As far as I'm aware, comments are not part of the standard JSON specification

That's true, but it doesn't prevent from showing comments in syntax.

mode-json.js
```
{
  token: "comment.start", // comments are not allowed, but who cares?
  regex: "\\/\\*",
  next: "comment"
}

```

> For comments, either [JSON5](https://github.com/ajaxorg/ace/blob/master/src/mode/json5.js) can be used, or JSONC could be added to the list of Ace modes

JSON5 does not provide syntax validation (now at least).

InspiredGuy:
I don't think we will go against the spec in JSON worker, and we're not looking for worker updates at the moment. I will close this PR.

There is a community project https://github.com/mkslanc/ace-linters, not owned by Ace, which provides language features for Ace (including syntax validation) and [it seems to support JSON5](https://github.com/mkslanc/ace-linters?tab=readme-ov-file#supported-languages). It might be helpful for your use case if you want comments.

PR #5517, opened 2024-03-21
fix: don't detach completion popup while typing when prefix is empty

akoreman:
*Issue #, if available:* NA

*Description of changes:* Currently, when auto showing completion during typing, the popup gets detached when the prefix is `""`. This can be problematic in situations when completions are coming in while typing sentences (e.g. when using AI code generation). When you pause after typing a space and a completion comes in from the completer, the popup will be detached because the prefix at that point in time is `""`.

Pull Request Checklist:

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -362,8 +362,6 @@ class Autocomplete {
             if (this.tooltipNode) {
                 this.updateDocTooltip();
             }
-        } else if (keepPopupPosition && !prefix && !inlineEnabled) {
-            this.detach();
         }
         this.changeTimer.cancel();
         this.observeLayoutChanges();
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -362,7 +362,7 @@ class Autocomplete {
             if (this.tooltipNode) {
                 this.updateDocTooltip();
             }
-        } else if (keepPopupPosition && !prefix) {
+        } else if (keepPopupPosition && !prefix && !inlineEnabled) {
             this.detach();
         }
         this.changeTimer.cancel();
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -749,6 +749,12 @@ Autocomplete.prototype.commands = {
         else
             return result;
     },
+    "Backspace": function(editor) {
+        editor.execCommand("backspace");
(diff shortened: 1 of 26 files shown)


src/autocomplete.js
```diff
@@ -362,8 +362,6 @@ class Autocomplete {
             if (this.tooltipNode) {
                 this.updateDocTooltip();
             }
-        } else if (keepPopupPosition && !prefix) {
-            this.detach();
```
andredcoliveira, line 366: Looks like this was added in https://github.com/ajaxorg/ace/commit/bd95f782d8fd5be28334e8783bba52406d95654c to close the popup when backspacing into an empty string. E.g.:

```
// type something like `con`
con
// press CTRL+Space -> pop-up renders with suggestions like "console.(...)"
// backspace 3 times
<empty line>
// what happens to the popup here after this PR?
```

akoreman: yeah it does change the behavior there

https://github.com/ajaxorg/ace/assets/34573235/b7d049df-2e1a-4691-931d-2feddd821a0c

akoreman: In some sense, backspacing until it becomes an empty prefix is similar to typing until it's an empty prefix, might be hard to keep the old behavior while backspacing while still allowing for empty prefix completion while typing ahead.

andredcoliveira: Maybe we could do something similar to the logic in https://github.com/ajaxorg/ace/blob/6e870fd08e06be47f9da971743d13b25c41f99c2/src/ext/language_tools.js#L140-L153?

I.e., different behaviour depending on whether it's a deletion or an insertion of text

andredcoliveira: Maybe a command for `"Backspace"` close to the other commands: https://github.com/ajaxorg/ace/blob/6e870fd08e06be47f9da971743d13b25c41f99c2/src/autocomplete.js#L738-L757

Something along the lines of "if backspace and no prefix then detach"

akoreman: I went for the second option in a new commit

andredcoliveira (approved):
Looks good to me :)

oykuyilmaz (approved):

PR #5519, opened 2024-03-21
fix: allow completers to trigger on newline characters

andredcoliveira:
*Issue #, if available:*
N/A

*Description of changes:*
Completers may want to use newline characters for triggering suggestions. This is particularly useful when showing suggestions after typing prompts in the form of inline comments in the case of GenAI suggestions.

Pull Request Checklist:
* [✓] No backwards incompatible changes were made to Ace's public interface which is defined through the main typings file (`ace.d.ts`) and its references:

### DIFF
--- a/src/edit_session.js
+++ b/src/edit_session.js
@@ -2367,18 +2367,17 @@ class EditSession {
     }
 
     /**
-     * @returns {string} the last character inserted into the editor
+     * @returns {string} the last character preceding the cursor in the editor
      */
-    getPreviousChar() {
+    getPrecedingCharacter() {
         var pos = this.selection.getCursor();
 
-        if (pos.row === 0) {
-            var line = this.getLine(pos.row);
-            var column = pos.column === 0 ? 0 : pos.column - 1;
-            return line[column];
+        if (pos.column === 0) {
+            return pos.row === 0 ? "" : this.doc.getNewLineCharacter();
         }
 
-        return this.doc.getNewLineCharacter();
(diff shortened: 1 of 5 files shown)


andredcoliveira:
> One of the public type files has been updated, plase make sure there are no backwards incompatible changes done in the PR.

Only a new method was added. No backwards incompatible changes

src/edit_session.js
```diff
@@ -2366,6 +2366,20 @@ class EditSession {
...
     }
 
+    /**
+     * @returns {string} the last character preceding the cursor in the editor
+     */
+    getPrecedingCharacter() {
+        var pos = this.selection.getCursor();
+
+        if (pos.column === 0) {
+            return pos.row === 0 ? "" : this.doc.getNewLineCharacter();
```
akoreman, line 2376: What happens if indentation is automatically inserted after pressing newline? For example when I have

```
function() {
```

and I press enter after the last bracket, Ace would automatically inserting indentation but I would probably still like to get the auto-triggered suggestion.

andredcoliveira: Great catch! I forgot to test that workflow after my second commit (https://github.com/ajaxorg/ace/pull/5519/commits/ae6b91aba9ef47e70cb536ea1ccc1e9cd578cda7). It wasn't triggering suggestions when indentation was automatically applied after a newline insertion.

Should be triggering correctly now with https://github.com/ajaxorg/ace/pull/5519/commits/a1ad3feb60f950281a595cd0bcc9674f6b2c4b72.

PR #5522, opened 2024-03-25
fix: apply class to autocomplete popup for no suggestions state

akoreman:
*Issue #, if available:* NA

*Description of changes:* Apply class when autocomplete popup is in the 'no suggestions' state to allow separate styling of that state.

Pull Request Checklist:

### DIFF
--- a/src/autocomplete_test.js
+++ b/src/autocomplete_test.js
@@ -403,6 +403,7 @@ module.exports = {
     },
     "test: trigger autocomplete for specific characters": function (done) {
         var editor = initEditor("document");
+        var newLineCharacter = editor.session.doc.getNewLineCharacter();
 
         editor.completers = [
             {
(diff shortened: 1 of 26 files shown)


src/autocomplete_test.js
```diff
@@ -462,12 +462,43 @@ module.exports = {
...
+        ];
+
+        user.type("app");
+
+        // Open autocompletion via key-binding and verify no empty message class
+        user.type("Ctrl-Space");
+        assert.equal(editor.completer.popup.isOpen, true);
+        assert.equal(editor.completer.popup.data[0].caption, "append");
+        assert.ok(!editor.completer.popup.renderer.container.classList.contains("ace_empty-message"));
+
```
whazor, line 499: It would be nice to also test the reverse `empty -> results`

akoreman: so trigger it 1 time without results (check class there) and then trigger with results and check class not there?

akoreman: Improved the test to test that scenario as well

src/autocomplete_test.js
```diff
@@ -462,12 +462,52 @@ module.exports = {
...
+                            caption: "append",
+                            value: "append"
+                        }
+                    ];
+                    callback(null, completions);
+                }
+            }
+        ];
+
+        user.type("b");
```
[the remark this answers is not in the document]
oykuyilmaz: I retract my comment, I've missed that you register the completer explicitly.

oykuyilmaz (approved):

Issue #5499, opened 2024-03-07
gutter: "Read annotations row X" aria-label doesn't contain the type of annotation


avinashbot:
### Describe the bug

In the gutter, when an annotation icon is present, the aria-label just states "Read annotations row X", but doesn't provide any information to assistive technology whether it's an info, warning, or error icon.

### Expected Behavior

Ideally, the aria-label should be something like "Error, read annotations, row 1" or "Warning, read annotations, row 1", where the string starts with the icon type of the button being focused.

### Current Behavior

When testing (with `enableKeyboardAccessibility` enabled), VoiceOver on Mac reads the icon as "Read annotations, row 1, button". No information about the icon itself is communicated to screen readers.

### Reproduction Steps

1. Open ace with a language that supports syntax checking (e.g. Javascript in the [kitchen sink](https://ace.c9.io/build/kitchen-sink.html))
2. Cause a syntax error or a warning.
3. See that the `.ace_gutter_annotation` in the DOM has an `aria-label` without the icon type mentioned in it.

### Additional Information/Context

Relevant lines: [src/layer/gutter.js](https://github.com/ajaxorg/ace/blob/d296fd4af374cb7dad41c437611b9da81761b780/src/layer/gutter.js#L445)

### Ace Version / Browser / OS / Keyboard layout

^1.32.6 / Any / Any / Qwerty

InspiredGuy:
Seems like this was resolved by https://github.com/ajaxorg/ace/pull/5524 back then, closing the issue

PR #5524, opened 2024-03-29
feat: Make translation system key based and add annotation type to gutter icon aria labels

akoreman:
*Issue #, if available:* #5499

*Description of changes:* Changing the API we have for UI string translations from being based on the default value of the string to a more abstract key. Aim was to do this in a way that doesn't break users who currently use translations for the default value of the strings.

Changes:
- Changes main translation identifier to a more abstract key
  - Falls back to old behavior if key is missing from provided translation file
- Updates the `nls` step in the build accordingly
- Updates existing translation files accordingly

Pull Request Checklist:


src/lib/app_config.js
```diff
@@ -142,14 +144,19 @@ class AppConfig {
...
      */
-    nls(string, params) {
-        if (messages && !messages[string])  {
-            warn("No message found for '" + string + "' in the provided messages, falling back to default English message.");
-        }
-        var translated = messages && messages[string] || string;
+    nls(key, defaultString, params) {
+        if (!messages[key])  {
+            warn("No message found for the key '" + key + "' in the provided messages, trying to find a translation for the default string '" + defaultString + "'.");
+            if (!messages[defaultString]) {
```
oykuyilmaz, line 154: are we keeping this default string logic for backwards compatibility reasons?

akoreman: yeah, I kept this so that Ace users who currently have translations set-up don't get those broken by this change (well that's the goal at least 😄)

Makefile.dryice.js
```diff
@@ -654,28 +654,34 @@ function extractCss(callback) {
...
-        var matches = text.match(/nls\s*\(\s*("([^"\\]|\\.)+"|'([^'\\]|\\.)+')/g);
+        var matches = text.match(/nls\s*\(\s*("([^"\\]|\\.)+"|'([^'\\]|\\.)+'),\s*("([^"\\]|\\.)+"|'([^'\\]|\\.)+')/g);
         matches && matches.forEach(function(m) {
-            var eng = m.replace(/^nls\s*\(\s*["']|["']$/g, "");
-            allMessages[eng] = "";
+            var match = m.match(/("([^"\\]|\\.)+"|'([^'\\]|\\.)+)/g);      
+            var key = match[0].replace(/["']|["']$/g, "");
+            var defaultString = match[1].replace(/["']|["']$/g, "");
+
+            // If the key not yet in the default file, add it:
```
oykuyilmaz, line 668: do we have a mechanism to keep "default_english_messages" up-to-date with the used keys?

oykuyilmaz: oh I see following line with `fs.writeFileSync`

oykuyilmaz (approved):

Makefile.dryice.js
akoreman: if you run `node makefile.dryice.js nls` the file will get updated with new keys, maybe we should include that in our build process but not sure where 🤔

Issue #5513, opened 2024-03-19
Suport soft-wrapping of ghost text


kevinushey:
### Describe the feature

Currently, when displaying ghost text within an Ace instance, that text is not soft-wrapped. For example observe how the first line (real text) is wrapped, while the second line is not:

<img width="627" alt="Screenshot 2024-03-18 at 6 40 09 PM" src="https://github.com/ajaxorg/ace/assets/1976582/cfc23bbb-7518-49e9-a5cd-f51e294f5f3e">

### Use Case

Soft-wrapping of ghost text when enabled would better match the user intention.

### Proposed Solution

It looks like ghost text is added as a single token "ghost_text" here:

https://github.com/ajaxorg/ace/blob/099011b0609d47c366cf3fcd09458723c314470c/src/virtual_renderer.js#L1761-L1762

I wonder if the ghost text could be split into separate tokens, and added individually? A simple implementation would be to just split on whitespace, and I think that would be enough to make sure the rendered text obeys soft-wrapping rules.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### ACE version used

version 1.32.7

kevinushey:
That's fantastic -- thank you!

PR #5540, opened 2024-04-26
Fix: Add soft wrap for ghost text

mkslanc:
*Issue #, if available:* #5513

*Description of changes:*
The current implementation wraps all long text based on the screen width. I am considering whether it makes sense to adhere to the user's wrap settings in this case. In my opinion, it would be beneficial to display and wrap the entire line of ghost text, but this point is open for discussion.

[screen-capture (58).webm](https://github.com/ajaxorg/ace/assets/32402726/b2428cd0-ea91-4e7a-9bc9-08584c204e0b)

**(Test completion includes sentences with newlines)**

Pull Request Checklist:

### DIFF
--- a/src/autocomplete/inline_test.js
+++ b/src/autocomplete/inline_test.js
@@ -101,8 +101,8 @@ module.exports = {
         inline.show(editor, completions[3], "f");
         editor.renderer.$loop._flush();
         assert.strictEqual(getAllLines(), textBase + "function foo() {");
-        assert.strictEqual(editor.renderer.$ghostTextWidget.text, "        console.log('test');\n    }");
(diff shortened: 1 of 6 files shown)


akoreman:
This looks a nice change! Do you see a nice way to communicate to users in the ghost text which lines breaks are due to the soft-wrapping and which line breaks are actual line breaks in the suggestion?

mkslanc:
> This looks a nice change! Do you see a nice way to communicate to users in the ghost text which line breaks are due to the soft-wrapping and which line breaks are actual line breaks in the suggestion?

I think I could make it work either by tweaking the CSS or by adding a wrapping symbol at the end of the wrapped lines. I'll test something and push the changes in time.

akoreman (approved):

akoreman:
Looks good to me, thanks for this contribution 🚀

PR #5560, opened 2024-05-22
feat: allow users to add arialabel to text input

akoreman:
*Issue #, if available:*

*Description of changes:* Currently, the aria-label of the text input says `Cursor at row x`, some users might want to add context about their editor to this aria label. This change adds an option, `textInputAriaLabel`, which is used to create and aria-label as `textInputAriaLabel, Cursor at row x`.

Pull Request Checklist:

### DIFF
--- a/ace.d.ts
+++ b/ace.d.ts
@@ -235,6 +235,7 @@ export namespace Ace {
     relativeLineNumbers: boolean;
     enableMultiselect: boolean;
     enableKeyboardAccessibility: boolean;
+    textInputAriaLabel: string;
   }
 
   export interface SearchOptions {
--- a/src/keyboard/textinput.js
+++ b/src/keyboard/textinput.js
@@ -90,7 +90,11 @@ TextInput= function(parentNode, host) {
             text.setAttribute("aria-roledescription", nls("text-input.aria-roledescription", "editor"));
             if(host.session) {
                 var row =  host.session.selection.cursor.row;
-                var arialLabel = `${host.$textInputAriaLabel}, ${nls("text-input.aria-label", "Cursor at row $0", [row + 1])}`;
+                var arialLabel = "";
+                if (host.$textInputAriaLabel) {
+                    arialLabel += `${host.$textInputAriaLabel}, `;
+                }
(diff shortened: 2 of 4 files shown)


ace.d.ts
```diff
@@ -235,6 +235,7 @@ export namespace Ace {
     relativeLineNumbers: boolean;
     enableMultiselect: boolean;
     enableKeyboardAccessibility: boolean;
+    textInputAriaLabel: string;
```
marinsokol5, line 238: So I guess nothing is optional here? Although everything is

src/keyboard/textinput.js
```diff
@@ -90,7 +90,8 @@ TextInput= function(parentNode, host) {
             text.setAttribute("aria-roledescription", nls("text-input.aria-roledescription", "editor"));
             if(host.session) {
                 var row =  host.session.selection.cursor.row;
-                text.setAttribute("aria-label", nls("text-input.aria-label", "Cursor at row $0", [row + 1]));
+                var arialLabel = `${host.$textInputAriaLabel}, ${nls("text-input.aria-label", "Cursor at row $0", [row + 1])}`;
```
marinsokol5, line 93: should we hide "," if there is textInputAriaLabel is not there? Basically to cater for the initial value.

src/keyboard/textinput_test.js
```diff
@@ -753,6 +753,18 @@ module.exports = {
         assert.equal(editor.getValue(), "");
         sendEvent("input", {key: {inputType: "historyRedo"}});
         assert.equal(editor.getValue(), "x");
+    },
+
+    "test: text input aria label": function() {
```
marinsokol5, line 758: Can we also add same test but without `textInputAriaLabel ` being set.

akoreman: good point, added

src/keyboard/textinput.js
akoreman: yeah that's fair, changed

ace.d.ts
marinsokol5: I see we have it in https://github.com/ajaxorg/ace/blob/master/ace.d.ts#L1104

marinsokol5 (approved):

Issue #5595, opened 2024-06-19
JSDoc `@param` not detected / wrong coloring


kungfooman:
### Describe the bug

JSDoc example:

```js
/**
 * Converts an amplitude spectrogram to the decibel scale. This computes `20 * log10(spectrogram / reference)`,
 * using basic logarithm properties for numerical stability. NOTE: Operates in-place.
 * 
 * The motivation behind applying the log function on the (mel) spectrogram is that humans do not hear loudness on a
 * linear scale. Generally to double the perceived volume of a sound we need to put 8 times as much energy into it.
 * This means that large variations in energy may not sound all that different if the sound is loud to begin with.
 * This compression operation makes the (mel) spectrogram features match more closely what humans actually hear.
 * 
 * @template {Float32Array|Float64Array} T
 * @param {T} spectrogram The input amplitude (mel) spectrogram.
 * @param {number} [reference=1.0] Sets the input spectrogram value that corresponds to 0 dB.
 * For example, use `np.max(spectrogram)` to set the loudest part to 0 dB. Must be greater than zero.
 * @param {number} [min_value=1e-5] The spectrogram will be clipped to this minimum value before conversion to decibels,
 * to avoid taking `log(0)`. The default of `1e-5` corresponds to a minimum of -100 dB. Must be greater than zero.
 * @param {number} [db_range=null] Sets the maximum dynamic range in decibels. For example, if `db_range = 80`, the
 * difference between the peak value and the smallest value will never be more than 80 dB. Must be greater than zero.
 * @returns {T} The modified spectrogram in decibels.
 */
function amplitude_to_db(spectrogram, reference = 1.0, min_value = 1e-5, db_range = null) {
    return _db_conversion_helper(spectrogram, 20.0, reference, min_value, db_range);
}
```

### Expected Behavior

Proper token coloring.

### Current Behavior

Missed `@param` token coloring:

![image](https://github.com/ajaxorg/ace/assets/5236548/ec722a0f-3f46-4a73-abe2-9743192fd5e7)

### Reproduction Steps

Just paste JSDoc example in any ACE instance configured for JavaScript.

### Possible Solution

You probably just missed the `-` in a RegExp.

### Ace Version / Browser / OS / Keyboard layout

All

PR #5596, opened 2024-06-20
Support JSDoc comment optional values with negative exponents

kungfooman:
Now we support JSDoc comment optional values with negative exponents like: `@param {number} [min_value=1e-5]`

Fixes: #5595 

I just added a minus to the `RegExp`, so now it is recognized as a proper token:

Before:

image

After:

image

### DIFF
--- a/src/mode/jsdoc_comment_highlight_rules.js
+++ b/src/mode/jsdoc_comment_highlight_rules.js
@@ -24,7 +24,7 @@ var JsDocCommentHighlightRules = function() {
                         ]
                     }, {
                         token: ["rparen.doc", "text.doc", "variable.parameter.doc", "lparen.doc", "variable.parameter.doc", "rparen.doc"],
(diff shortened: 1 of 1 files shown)


nightwing (commented):
LGTM

But do we need to consider more possibilities here, like `[a="string"]` or `[a = 1]`?

kungfooman:
> But do we need to consider more possibilities here, like `[a="string"]` or `[a = 1]`?

Definitely good point, once we have `"` it has to be nearly fully parsed... :thinking: 

Consider:

```js
/**
 * @param {number} [reference="@param {}"] Sets the input spectrogram value that corresponds to 0 dB.
 */
```

Maybe that goes a bit too far... at least I didn't come across such JSDoc in real life yet.

kungfooman:
Thank you again @nightwing, now it's supporting both abilities you mentioned plus `'` aswell:

image

nightwing (approved):

nightwing:
thank you!

Issue #5589, opened 2024-06-12
latex: (bright) GitHub theme only black and with only a few commands highlighted


dbitouze:
### Describe the bug

On https://ace.c9.io/build/kitchen-sink.html, with a LaTeX document, the (bright) GitHub theme displays LaTeX code only in black with only a few commands highlighted (in bold face).

### Expected Behavior

Same kind of behavior as the github.com site or, for instance, the Ace GitHub Dark theme which displays LaTeX code in colors, with all of the command names highlighted.

### Current Behavior

In the following default LaTeX code provided by the Kitchen Sink, only `\usepackage` is highlighted by the Ace (bright) GitHub theme:
```latex
\usepackage{amsmath}
\title{\LaTeX}
\date{}
\begin{document}
  \maketitle
  \LaTeX{} is a document preparation system for the \TeX{}
  typesetting program. It offers programmable desktop publishing
  features and extensive facilities for automating most aspects of
  typesetting and desktop publishing, including numbering and
  cross-referencing, tables and figures, page layout, bibliographies,
  and much more. \LaTeX{} was originally written in 1984 by Leslie
  Lamport and has become the dominant method for using \TeX; few
  people write in plain \TeX{} anymore. The current version  is
  \LaTeXe.
 
  % This is a comment; it will not be shown in the final output.
  % The following shows a little of the typesetting power of LaTeX:
  \begin{align}
    E &= mc^2                              \\
    m &= \frac{m_0}{\sqrt{1-\frac{v^2}{c^2}}}
  \end{align}
\end{document}
```
As you can see above if you're using the default bright theme of this github.com site, the LaTeX code is colorful.

### Reproduction Steps

1. Go to https://ace.c9.io/build/kitchen-sink.html.
2. Choose a LaTeX document.
3. Choose the Bright GitHub theme.

### Ace Version / Browser / OS / Keyboard layout

- **Ace version:** the latest one, I guess.
- **Browser:** Mozilla/5.0 (X11; Linux x86_64; rv:126.0) Gecko/20100101 Firefox/126.0.
- **OS:** Linux x86_64. 
- **Keyboard:** azerty-fr.

InspiredGuy:
It seems like this is an issue with other document types as well, I did check couple of others (Kotlin, Go, Javascript), and there is also lack of highlighting for some syntax types.

akoreman:
Should be fixed by #5597 which will be included in the next Ace release.

dbitouze:
[...]

PR #5597, opened 2024-06-20
Port modern github light theme

mkslanc:
*Issue #, if available:* #5589

*Description of changes:*
- I couldn't locate the source of the original GitHub theme, so I have not altered it to avoid disrupting the user experience for those accustomed to it. I have marked it as `legacy` for clarity.
- This update is a port of the official [GitHub theme from VS Code](https://marketplace.visualstudio.com/items?itemName=GitHub.github-vscode-theme). The port was inspired by the [ace's tmTheme tool](https://github.com/ajaxorg/ace/blob/master/tool/tmtheme.js)

image
The differences in colors are due to the usage of different tokens in the highlighting modes.

**P.S.: If this proves valuable, I could also port the remaining dark and high contrast themes.**

Pull Request Checklist:


akoreman:
Hi, I think the new theme looks pretty good! Not fully sure whether the `classic` in the theme name conveys that this is the most modern one people should use if they want to have a GH-like theme? Maybe `GitHub Default` could work (mirroring how they're called in the VSC extension)? This could also help to make it more clear which theme is which if we want to add the high-contrast theme variants in future as well.

mkslanc:
> Hi, I think the new theme looks pretty good! Not fully sure whether the `classic` in the theme name conveys that this is the most modern one people should use if they want to have a GH-like theme? Maybe `GitHub Default` could work (mirroring how they're called in the VSC extension)? This could also help to make it more clear which theme is which if we want to add the high-contrast theme variants in future as well.

Hey, I agree. + I added few missed token scopes for html/xml

akoreman (approved):

dbitouze:
I can't wait to see this theme integrated into Ace! :blush:

Issue #5575, opened 2024-06-03
(default_gutter_handler): showTooltip is CSP incompliant - assigns innerHTML


dimovpetar:
### Describe the bug

GutterTooltip#showTooltip uses `this.setHtml`, which results in `innerHTML=...`
https://github.com/ajaxorg/ace/blob/a3277eaef554bdf6393e3957d60f5420e1d7130f/src/mouse/default_gutter_handler.js#L227

### Expected Behavior

Tooltip is created in CSP compliant way

### Current Behavior

Tooltip is created in CSP incompliant way

### Reproduction Steps

1. Open any ace editor and type something invalid
2. Hover on the tooltip with the mouse

### Ace Version / Browser / OS / Keyboard layout

1.31.1

dimovpetar:
@akoreman thanks for the fix. Is this issue ready to be closed now?

akoreman:
Yeah this issue should be resolved now

PR #5601, opened 2024-06-27
fix: use dom builder for gutter tooltip and inline widget

akoreman:
*Issue #, if available:* #5575

*Description of changes:* Changes how we construct the gutter tooltip and inline widget to use the dom building utilities.

Pull Request Checklist:


whazor (approved):

src/virtual_renderer_test.js
```diff
@@ -395,7 +395,7 @@ module.exports = {
         editor.renderer.$loop._flush();
         assert.equal(editor.renderer.content.textContent, "abcdefThis is a long test text that is longer than ");
 
-        assert.equal(editor.session.lineWidgets[0].el.innerHTML, "<div class=\"ghost_text_line_wrapped\">30 characters</div><div></div><div>Ghost3</div>");
+        assert.equal(editor.session.lineWidgets[0].el.innerHTML, `<div class="ghost_text_line_wrapped">30 characters</div><div class=""></div><div class="">Ghost3</div>`);
```
andredcoliveira, line 398: Where do the empty classes come from? Are they a consequence of using `dom.createElement`?

andredcoliveira: Could you confirm whether this was tested manually (e.g., in the kitchen sink) too?

andredcoliveira (approved):
Nice

src/virtual_renderer_test.js
akoreman: the empty classes are a consequence from ` chunkDiv.className = el.wrapped ? "ghost_text_line_wrapped": "";`. I can change it so we only set the classname if it's needed that might be cleaner indeed.

akoreman: > Could you confirm whether this was tested manually (e.g., in the kitchen sink) too?

yes 

Screenshot 2024-06-27 at 11 22 21

andredcoliveira (approved):

PR #5611, opened 2024-07-13
move default gutter png icons rendering to .ace_icon

InspiredGuy:
*Issue #, if available:*
https://github.com/ajaxorg/ace/issues/5380

*Description of changes:*
Moving the background which sets the gutter png icons from  `.ace_gutter-cell.ace_error` to `.ace_gutter-cell.ace_error .ace_icon`. It's the same container used for displaying annotations in folded block.

Initially I've tested a naive fix for this by moving the background position to top instead of center (same positioning would be achived with `background-position: 2px -1px;` instead of `background-position: 2px center;`). It looked OK on default font size, but with font size changing it did not keep up the alignment with line numbers (see screenshot below). Solution implemented in this PR keeps the icons aligned with line numbers regardless of font size.
image

Pull Request Checklist:

### DIFF
--- a/src/css/editor-css.js
+++ b/src/css/editor-css.js
@@ -130,24 +130,24 @@ module.exports = `
     float: left;
 }
 
-.ace_gutter-cell.ace_error, .ace_icon.ace_error, .ace_icon.ace_error_fold {
+.ace_gutter-cell.ace_error .ace_icon, .ace_icon.ace_error, .ace_icon.ace_error_fold {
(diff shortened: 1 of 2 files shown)


InspiredGuy:
@nightwing @akoreman let me know if you have any concerns about moving the background to another element. I've checked the codebase but it doesn't seem the target element is used for anything except for showing errors in folded block (which is not affected by this change).

akoreman:
I think the one concern would be what would happen if people currently have a custom theme theme set the background image to `.ace_gutter-cell.ace_error`? Would this change break them by displaying both background images on top of each other?

On the other hand, this is a bullet we might need to bite at some point and accept that for these rare use cases people will have to make a change, but in that case, would it be better to just fully deprecate the PNG icons and fully go for the new SVG ones?

InspiredGuy:
> I think the one concern would be what would happen if people currently have a custom theme theme set the background image to .ace_gutter-cell.ace_error? Would this change break them by displaying both background images on top of each other?

Yeah, it's not like their app won't run but this scenario you've described is certainly possible. It might be even worse since it will be implicit.

> On the other hand, this is a bullet we might need to bite at some point and accept that for these rare use cases people will have to make a change, but in that case, would it be better to just fully deprecate the PNG icons and fully go for the new SVG ones?

This would be breaking on the interface level as we'd need to deal with the API property we have. If we remove it it's kind of a breaking change, and if we just ignore the passed value it's also kind of a breaking change. It might be wise to save such moves until the next major version :)

InspiredGuy:
I've updated the PR to calculate the background position based on the font size and icon size in the currently used container. Value for line height is 1.2, which is the same as `normal` (see [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values)). Icon size is 16x16 implicitly. It can be tested here - https://raw.githack.com/ajaxorg/ace/fix-gutter-icon-position/kitchen-sink.html

Not the prettiest solution, and if we think it's too risky I'm OK to discard it completely, this issue is not super important.

FYI @akoreman

InspiredGuy:
@nightwing I've checked the codebase and it seems like css variables are not used anywhere yet. Is there a reason for this? They seem well-supported across browsers.

PR #5627, opened 2024-08-18
Hint and security annotations

andredcoliveira:
*Issue #, if available:*
- N/A

*Description of changes:*
- Adds a `prettier` config, which is useful for prettifying modified lines via your editor of choice
  - Does not add `prettier` itself as a dependency
- Adds _Hint_ gutter annotation (called _Suggestion_ in user interfaces)
  - Part of the LSP protocol
  - Copies the color scheme of _Information_ (at least for now, to ensure accessible colours)
  - Defaults to the _Information_ icon when `useSvgGutterIcons` isn't set
  - Icon: https://github.com/cloudscape-design/components/blob/e7561bf039ccec3f97a0d930eaa944383254169f/src/icon/icons/suggestions.svg
- Adds _Security_ gutter annotation & inline marker
  - Not part of the LSP protocol, but also useful
  - Copies the color scheme of _Warning_ (at least for now, to ensure accessible colours)
  - Defaults to the _Warning_ icon when `useSvgGutterIcons` isn't set
  - Icon: https://github.com/cloudscape-design/components/blob/e7561bf039ccec3f97a0d930eaa944383254169f/src/icon/icons/security.svg

SVG HTML code:

```html
<!-- Hint -->
<svg viewBox="0 0 20 16" xmlns="http://www.w3.org/2000/svg">
    <g stroke-width="2" stroke="blue" fill="none" shape-rendering="geometricPrecision">
        <path class="stroke-linejoin-round" d="M6 14H10"/>
        <path d="M8 11H9C9 9.47002 12 8.54002 12 5.76002C12.02 4.40002 11.39 3.36002 10.43 2.67002C9 1.64002 7.00001 1.64002 5.57001 2.67002C4.61001 3.36002 3.98 4.40002 4 5.76002C4 8.54002 7.00001 9.47002 7.00001 11H8Z"/>
    </g>
</svg>

<!-- Security -->
<svg viewBox="0 0 20 16" xmlns="http://www.w3.org/2000/svg">
    <g stroke-width="2" stroke="darkorange" fill="none" shape-rendering="geometricPrecision">
        <path class="stroke-linejoin-round" d="M8 14.8307C8 14.8307 2 12.9047 2 8.08992V3.26548C5.31 3.26548 7.98999 1.34918 7.98999 1.34918C7.98999 1.34918 10.69 3.26548 14 3.26548V8.08992C14 12.9047 8 14.8307 8 14.8307Z"/>
        <path d="M2 8.08992V3.26548C5.31 3.26548 7.98999 1.34918 7.98999 1.34918"/>
        <path d="M13.99 8.08992V3.26548C10.68 3.26548 8 1.34918 8 1.34918"/>
        <path class="stroke-linejoin-round" d="M8 4V9"/>
        <path class="stroke-linejoin-round" d="M8 10V12"/>
    </g>
</svg>
```

Pull Request Checklist:


src/css/editor-css.js
```diff
@@ -163,11 +168,19 @@ module.exports = `
...
     background-color: crimson;
 }
+.ace_icon_svg.ace_security_fold {
```
akoreman, line 180: As far as I can tell, this is the same SVG as for the non-fold version right? Are you planning to add a folded version of the icon in a follow up?

andredcoliveira: Correct, it's the same icon. Not sure yet. I plan to either a) get a folded version and update this PR, b) remove security from being in folds (probably downgrade "below warning" too if this happens) or c) proceed with this PR and have a follow-up item.

The outcome is pending some discussions outside of this PR.

src/mouse/default_gutter_handler.js
```diff
@@ -151,18 +151,27 @@ class GutterTooltip extends Tooltip {
...
     
-    static get annotationLabels() { return {
+    static get annotationLabels() {
+        return {
             error: {
-                singular: nls("gutter-tooltip.aria-label.error.singular", "error"), 
+                singular: nls("gutter-tooltip.aria-label.error.singular", "error"),
                 plural: nls("gutter-tooltip.aria-label.error.plural", "errors")
             },
+            security: {
```
akoreman, line 160: We should also add these new strings to https://github.com/ajaxorg/ace/blob/master/src/lib/default_english_messages.js#L44

I think that should work automatically if you run `node Makefile.dryice.js nls`

src/lib/default_english_messages.js
```diff
@@ -38,7 +38,13 @@ var defaultEnglishMessages = {
     "gutter-tooltip.aria-label.warning.singular": "warning",
     "gutter-tooltip.aria-label.warning.plural": "warnings",
     "gutter-tooltip.aria-label.info.singular": "information message",
-    "gutter-tooltip.aria-label.info.plural": "information messages"
+    "gutter-tooltip.aria-label.info.plural": "information messages",
+    "gutter.annotation.aria-label.security": "Read annotations row $0",
```
akoreman, line 42: Like we have here for `error`, `warning`, `info`, these strings should include the severity of the annotation (so they should be something like `Security finding, read annotations row $0`). Probably best to change the string in `gutter.js`.

src/lib/default_english_messages.js
```diff
@@ -38,7 +38,13 @@ var defaultEnglishMessages = {
...
+    "gutter.annotation.aria-label.hint": "Read annotations row $0",
```
akoreman, line 43: Same as above

akoreman (approved):

PR #5630, opened 2024-08-21
feat: allow setting marker type for MarkerGroups

akoreman:
*Issue #, if available:* NA

*Description of changes:* When using `MarkerGroup`s, currently, all markers default to being of the `text` type in which only the text itself is highlighted within a range. Users of `MarkerGroup`s might want to have different types of markers. 

This change adds an `options` parameter to allow setting marker type to `fullLine` or `line`. In `fullLine`, all the rows within the range will be fully highlighted, for `line` we highlight the same as default but when newlines are part of the range we fill up the highlight from the last character within the line to the right side of the editor.

Pull Request Checklist:

### DIFF
--- a/src/marker_group.js
+++ b/src/marker_group.js
@@ -16,7 +16,7 @@ Potential improvements:
 class MarkerGroup {
     /**
      * @param {EditSession} session
-     * @param {{markerType: "fullLine" | "line"}} options
+     * @param {{markerType: "fullLine" | "line" | undefined}} [options]
      */
     constructor(session, options) {
         if (options)
--- a/src/marker_group.js
+++ b/src/marker_group.js
@@ -16,7 +16,11 @@ Potential improvements:
 class MarkerGroup {
     /**
      * @param {EditSession} session
-     * @param {{markerType: "fullLine" | "line" | undefined}} [options]
+     * @param {{markerType: "fullLine" | "line" | undefined}} [options] Options controlling the behvaiour of the marker.
+     * User `markerType` to control how the markers which are part of this group will be rendered:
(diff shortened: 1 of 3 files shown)


src/marker_group.js
```diff
@@ -59,12 +64,12 @@ class MarkerGroup {
...
      * @param {Marker} markerLayer
      * @param {EditSession} session
-     * @param {{ firstRow: any; lastRow: any; }} config
+     * @param {LayerConfig} config
      */
     update(html, markerLayer, session, config) {
         if (!this.markers || !this.markers.length)
             return;
-        var visibleRangeStartRow = config.firstRow, visibleRangeEndRow = config.lastRow;
+        var visibleRangeStartRow = config.firstRow, visibleRangeEndRow = 100;
```
whazor, line 72: should this be changed?

akoreman: ah whoops, that was just a change I made while playing around thanks for catching that 😅

src/marker_group.js
```diff
@@ -15,8 +16,12 @@ Potential improvements:
 class MarkerGroup {
     /**
      * @param {EditSession} session
+     * @param {{markerType: "fullLine" | "line" | undefined}} [options]
```
whazor, line 19: could you describe here what each argument does?

src/marker_group.js
akoreman: Removed that change

src/marker_group.js
akoreman: added a JSDoc comment

whazor (approved):

PR #5665, opened 2024-11-04
fix(a11y): update aria-label of textinput on cursor move

babalaui:
*Issue #, if available:*

*Description of changes:* The aria-label on ace text-input ("Cursor at row X"), was being set only on focus, which means that if the cursor moved within the editor, the label would still be set for the initial row. This lead to the screen reader announcing the wrong row number when accepting a suggestion.

With this change, the aria-label will be updated on every cursor change.

Pull Request Checklist:

### DIFF
--- a/src/keyboard/textinput.js
+++ b/src/keyboard/textinput.js
@@ -73,6 +73,19 @@ TextInput= function(parentNode, host) {
         
         numberOfExtraLines = number;
     };
+
+    this.setAriaLabel = function() {
+        var ariaLabel = "";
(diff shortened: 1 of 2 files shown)


akoreman:
Change looks good to me, did you verify that the screen reader doesn't read the updated aria label on every cursor change?

babalaui:
@akoreman yep, verified on mac os + voiceover, screen reader still only reads it when focusing editor.

akoreman (approved):

PR #5675, opened 2024-11-15
feat: Multiline Search Support: line breaks `\n`

sekedus:
*Issue #2869*

*Description of changes:*

* highlight line break `\n`
* search line break `\n`
* replace line break `\n`
* replace **{search}** value with `\n` or `\t`
ㅤ

### Benefit

Users can now perform search/replace line breaks (`\n`) with regex.
ㅤ

### Test

1. Open: https://sekedus.github.io/ace/kitchen-sink.html
2. Settings: `Document: JavaScript`
3. Set value:
  ```javascript
  /**
   * 
   * @param {string[]} items
   * @param nada
   */
  function foo(items, nada) {
      for (var i=0; i<items.length; i++) {
          alert(items[i] + "juhu\n");
      }	/* Real Tab */
  
  
  
  
  }
  
  
  // test search/replace line break with regexp

				

  ```
4. Open **search box**
    * Fill in search field with `\n` or `\n+` or `\n\s+` or `\)\s\{\n\s+`
    * Enable **RegExp mode**

   screenshot-ace-searchbox-2024_11_15-20_18_39
5. **Example**: 
    * Replace two or more line breaks with one line break:
      - `\n{2}` with `\n`
      - `\n+` with `\n`
    * Remove line breaks, replace `\n` with `{empty}`
    * Replace `{any_string}` with `\n` or`\t`
ㅤ

Pull Request Checklist:

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -617,10 +617,6 @@ class Editor {
 
         // Update cursor because tab characters can influence the cursor position.
         this.$cursorChange();
-
-        // Updates "updateCounter" and "ace_nomatch" in the search box
-        if (this.searchBox && this.searchBox.active === true)
-            this.searchBox.find(false, false, true);
     }
 
     /**
--- a/src/editor.js
+++ b/src/editor.js
@@ -617,6 +617,10 @@ class Editor {
 
         // Update cursor because tab characters can influence the cursor position.
         this.$cursorChange();
+
+        // Updates "updateCounter" and "ace_nomatch" in the search box
+        if (this.searchBox && this.searchBox.active === true)
+            this.searchBox.find(false, false, true);
     }
 
     /**
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -293,6 +293,9 @@ class SearchBox {
         this.element.style.display = "";
         this.replaceOption.checked = isReplace;
 
+        if (this.editor.$search.$options.regExp)
+            value = lang.escapeRegExp(value);
+
         if (value)
             this.searchInput.value = value;
 
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -51,6 +51,7 @@ class SearchBox {
         this.element = div.firstChild;
 
         this.setSession = this.setSession.bind(this);
+        this.$onEditorInput = this.onEditorInput.bind(this);
 
         this.$init();
         this.setEditor(editor);
@@ -70,6 +71,11 @@ class SearchBox {
     setSession(e) {
         this.searchRange = null;
         this.$syncOptions(true);
+    }
+
+    // Auto update "updateCounter" and "ace_nomatch"
+    onEditorInput() {
+        this.find(false, false, true);
     }
 
     /**
@@ -283,6 +289,7 @@ class SearchBox {
         this.active = false;
         this.setSearchRange(null);
         this.editor.off("changeSession", this.setSession);
+        this.editor.off("input", this.$onEditorInput);
 
         this.element.style.display = "none";
         this.editor.keyBinding.removeKeyboardHandler(this.$closeSearchBarKb);
@@ -296,6 +303,7 @@ class SearchBox {
     show(value, isReplace) {
         this.active = true;
         this.editor.on("changeSession", this.setSession);
+        this.editor.on("input", this.$onEditorInput);
         this.element.style.display = "";
         this.replaceOption.checked = isReplace;
 
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -218,8 +218,10 @@ class SearchBox {
              * Convert all line ending variations to Unix-style = \n
              * Windows (\r\n), MacOS Classic (\r), and Unix (\n)
              */
-            if (editor.$search.$isMultilineSearch(editor.getLastSearchOptions()))
+            if (editor.$search.$isMultilineSearch(editor.getLastSearchOptions())) {
                 value = value.replace(/\r\n|\r|\n/g, "\n");
+                editor.session.doc.$autoNewLine = "\n";
+            }
 
             var offset = editor.session.doc.positionToIndex(editor.selection.anchor);
             if (this.searchRange)
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -214,7 +214,7 @@ class SearchBox {
                 ? editor.session.getTextRange(this.searchRange)
                 : editor.getValue();
 
-            if (editor.$search.$isMultilineSearch(editor.$search.$options))
+            if (editor.$search.$isMultilineSearch(editor.getLastSearchOptions()))
                 value = value.replace(/\r\n|\r|\n/g, "\n");
 
             var offset = editor.session.doc.positionToIndex(editor.selection.anchor);
--- a/src/ext/searchbox.js
+++ b/src/ext/searchbox.js
@@ -214,6 +214,9 @@ class SearchBox {
                 ? editor.session.getTextRange(this.searchRange)
                 : editor.getValue();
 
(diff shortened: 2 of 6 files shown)


sekedus:
> Does this also resolve #5674?

ㅤ
Yes, its also resolve #5674

screenshot-ace-line-breaks-2024_11_18-18_36_36

src/search.js
```diff
@@ -248,13 +326,86 @@ class Search {
         return re;
     }
 
+    $isMultilineSearch(options) {
+        return options.re && /\\r\\n|\\r|\\n/.test(options.re.source) && options.regExp && !options.$isMultiLine;
```
nlujjawal, line 330: options.$isMultiLine should be true in case if search contains \n.
Reference: https://github.com/ajaxorg/ace/blob/5e1e524d7ef04e13291b5a979fb4166e973e61f0/ace-internal.d.ts#L667C32-L667C34

src/search.js
```diff
@@ -248,13 +326,86 @@ class Search {
...
+                if (row > last)
+                    break;
+                var next = session.getLine(row++);
+                line = line == null ? next : line + "\n" + next;
+            }
+
+            var match = re.exec(line);
+            re.lastIndex = 0;
+            if (match) {
+                var before = line.slice(0, match.index).split("\n");
```
nlujjawal, line 348: Can we rewrite the below 2 lines like
    // Split text before match and the match itself into lines
    const beforeMatch = line.slice(0, match.index).split("\n");
    const matchedText = match[0].split("\n");

src/search.js
```diff
@@ -322,43 +473,61 @@ class Search {
...
-                    }
-                    if (m.index + length > endIndex)
-                        break;
-                    matches.push(m.index, length);
-                }
-                for (var i = matches.length - 1; i >= 0; i -= 2) {
-                    var column = matches[i - 1];
-                    var length = matches[i];
-                    if (callback(row, column, row, column + length))
+                if (multiline) {
```
nlujjawal, line 476: should we reset regex lastIndex above this line just to reset it?
 re.lastIndex = 0;

src/search_highlight.js
```diff
@@ -35,16 +35,27 @@ class SearchHighlight {
...
             if (ranges == null) {
-                ranges = lang.getMatchOffsets(session.getLine(i), this.regExp);
-                if (ranges.length > this.MAX_RANGES)
-                    ranges = ranges.slice(0, this.MAX_RANGES);
-                ranges = ranges.map(function(match) {
-                    return new Range(i, match.offset, i, match.offset + match.length);
-                });
+                if (multiline) {
+                    ranges = [];
+                    var matches = _search.$multiLineForward(session, this.regExp, i, end, true);
```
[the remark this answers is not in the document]
sekedus: ACK.

src/ext/searchbox.js
```diff
@@ -214,6 +214,9 @@ class SearchBox {
                 ? editor.session.getTextRange(this.searchRange)
                 : editor.getValue();
 
+            if (editor.$search.$isMultilineSearch(editor.getLastSearchOptions()))
+                value = value.replace(/\r\n|\r|\n/g, "\n");
```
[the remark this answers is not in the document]
sekedus: ACK.

src/search.js
```diff
@@ -116,11 +116,17 @@ class Search {
...
         } else {
-            for (var i = 0; i < lines.length; i++) {
-                var matches = lang.getMatchOffsets(lines[i], re);
-                for (var j = 0; j < matches.length; j++) {
-                    var match = matches[j];
-                    ranges.push(new Range(i, match.offset, i, match.offset + match.length));
+            for (var matches, i = 0; i < lines.length; i++) {
+                if (this.$isMultilineSearch(options)) {
+                    matches = this.$multiLineForward(session, re, i, lines.length);
+                    ranges.push(new Range(matches.startRow, matches.startCol, matches.endRow, matches.endCol));
```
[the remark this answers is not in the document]
sekedus: ACK.

src/search.js
```diff
@@ -166,11 +240,15 @@ class Search {
         if (!re)
             return;
 
+        if (this.$isMultilineSearch(options))
```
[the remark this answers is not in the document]
sekedus: ACK.

src/search.js
sekedus: ACK.

src/search.js
sekedus: Yes, see: https://stackoverflow.com/a/4724920/7598333

sekedus: Yes, see: https://stackoverflow.com/a/4724920

src/search.js
```diff
@@ -508,8 +516,11 @@ class Search {
         else {
             var forEachInLine = function(row, startIndex, callback) {
                 re.lastIndex = startIndex;
-                if (multiline) {
+                if (mtSearch) {
                     var pos = mtForward(session, re, row, lastRow);
```
nlujjawal, line 520: can we keep calling function consistent earlier its been called with this.$multiLineForward on line 121.

src/search.js
nlujjawal: So can we add
re.lastIndex = 0;

src/search.js
```diff
@@ -116,9 +116,12 @@ class Search {
                     row = row + len - 2;
             }
         } else {
-            for (var matches, i = 0; i < lines.length; i++) {
+            for (var matches, i = 0, lng = lines.length; i < lng; i++) {
                 if (this.$isMultilineSearch(options)) {
-                    matches = this.$multiLineForward(session, re, i, lines.length);
+                    matches = this.$multiLineForward(session, re, i, lng);
+                    var end_row = matches.endRow <= lng ? matches.endRow - 1 : lng;
```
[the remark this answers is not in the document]
sekedus: ACK

src/search.js
sekedus: No, because the parameter is different.

src/search.js
sekedus: Sorry, I wasn’t focused earlier.

No need, because it’s already set in `multiLineBackwardMatch`.

nlujjawal (approved):

sekedus:
> Hey @sekedus Do you have time for writing test cases for the changes as well?

Done.

nlujjawal:
@sekedus some test case requirement is failing will you be able to resolve this?

sekedus:
> @sekedus some test case requirement is failing will you be able to resolve this?

Sorry, I don't know where the error is and how to resolve it.

Any suggestions?

src/editor.js
```diff
@@ -617,6 +617,10 @@ class Editor {
 
         // Update cursor because tab characters can influence the cursor position.
         this.$cursorChange();
+
+        // Updates "updateCounter" and "ace_nomatch" in the search box
+        if (this.searchBox && this.searchBox.active === true)
+            this.searchBox.find(false, false, true);
```
nightwing, line 623: Could you please move this to an input event handler in searchbox, something like `editor.on('input', this.$onEditorInput)` and remove the listener when deactivating. `onDocumentChange` can  be called multiple times during one edit, and input event debounces that with a timeout.
 Also in one of tests add a scenario of changing editor value, e.g. 
```js
editor.insert("searchedText") 
setTimeout(function() {
 // verify that search counter have been updated
}, 50)
```
to prevent coverlay complaining about untested lines

nightwing (approved):

nlujjawal:
@sekedus The failure of checks is due to some lines of code not having test cases.
You can view in this link about those lines https://github.com/ajaxorg/ace/pull/5675/checks?check_run_id=33534340733.

akoreman (approved):

PR #5691, opened 2024-12-06
WIP: Improve types

mkslanc:
*Issue #, if available:*

*Description of changes:*

Pull Request Checklist:

### DIFF
--- a/ace-internal.d.ts
+++ b/ace-internal.d.ts
@@ -308,8 +308,8 @@ export namespace Ace {
         coverGutter?: boolean,
         pixelHeight?: number,
         $fold?: Fold,
-        type?:any,
-        destroy?:()=>void;
+        type?: any,
+        destroy?: () => void;
         coverLine?: boolean,
         fixedWidth?: boolean,
         fullWidth?: boolean,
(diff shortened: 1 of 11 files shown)


akoreman:
Hey @mkslanc, thanks for the PR! I think it's a step in the right direction, is there anything blocking it from being ready for review?

mkslanc:
Hey @akoreman . I just waited until #5690 was merges, as it has some type intersections with that PR.

akoreman (approved):

ace-internal.d.ts
```diff
@@ -1554,7 +1564,3 @@ declare module "./src/mouse/default_gutter_handler" {
     }
 }
 
-declare module "./src/lib/keys" {
```
akoreman, line 1557: Missed this during the initial review, what was the reason for removing the type declaration of `lib/keys`?

mkslanc: Actually that was wrong, It exported only one function without repeating structure for this module, which is wild mix of 
```javascript
oop.mixin(exports, Keys);
exports.default = exports;
exports.keyCodeToString
```

PR #5699, opened 2024-12-17
chore: allow custom version number release script

akoreman:
*Issue #, if available:* NA

*Description of changes:* Currently, when releasing a new Ace version, the new version number is automatically picked up through conventional commits. Sometimes we want to be able to override this.

Pull Request Checklist:

### DIFF
--- a/tool/release.sh
+++ b/tool/release.sh
@@ -56,7 +56,13 @@ git checkout -- package.json
 git checkout -- CHANGELOG.md
 npm run changelog
 VERSION_NUM="$(node -p "require('./package.json').version")";
-echo "version number for the build is" $VERSION_NUM
+echo "recommended version number for the build is" $VERSION_NUM 
+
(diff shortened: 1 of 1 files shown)


InspiredGuy (approved):

tool/release.sh
```diff
@@ -56,7 +56,13 @@ git checkout -- package.json
 git checkout -- CHANGELOG.md
 npm run changelog
 VERSION_NUM="$(node -p "require('./package.json').version")";
-echo "version number for the build is" $VERSION_NUM
+echo "recommended version number for the build is" $VERSION_NUM 
+
+read -p "do you want to continue with the recommended version number? [y/n] " yn
```
whazor, line 61: seems like the default is `y` but not mentioned.

whazor: ahh, you can ignore this, seems like the rest of the file also uses that pattern

whazor (approved):

PR #5701, opened 2024-12-20
fix: hide gutter tooltip event missing from editor signal

nlujjawal:
*Description of changes:*
hide gutter tooltip event is not getting fired from editor signal in case when the cursor moves to a non-active gutter row which hides the active gutter tooltip.

Pull Request Checklist:


src/mouse/default_gutter_handler.js
```diff
@@ -271,6 +271,9 @@ class GutterTooltip extends Tooltip {
     }
 
     hideTooltip() {
+        if(!this.$element){
```
akoreman, line 274: Could it make sense to use `this.isOpen` here instead? That way we centralize how we check whether the tooltip is currently shown

(https://github.com/ajaxorg/ace/blob/master/src/tooltip.js#L20)

akoreman (approved):

PR #5703, opened 2024-12-20
expose gutter tooltip row information in gutter tooltip class

nlujjawal:
*Issue #, if available:*

*Description of changes:*
Exposing gutter tooltip row information in gutter tooltip class which can be used to get which rows tooltip is visible currently.

Pull Request Checklist:

### DIFF
--- a/src/mouse/default_gutter_handler.js
+++ b/src/mouse/default_gutter_handler.js
@@ -133,6 +133,8 @@ class GutterTooltip extends Tooltip {
     constructor(editor) {
         super(editor.container);
         this.editor = editor;
+        /**@type {Number | Undefined}*/
+        this.visibleTooltipRow;
     }
 
     setPosition(x, y) {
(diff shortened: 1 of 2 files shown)


src/mouse/default_gutter_handler.js
```diff
@@ -133,6 +133,7 @@ class GutterTooltip extends Tooltip {
     constructor(editor) {
         super(editor.container);
         this.editor = editor;
+        this.visibleToolTipRow;
```
akoreman, line 136: nitpick: in the rest of the file we don't capitalize the second T in Tooltip, so we might want to be consistent here as well

src/mouse/default_gutter_handler.js
```diff
@@ -133,6 +133,10 @@ class GutterTooltip extends Tooltip {
     constructor(editor) {
         super(editor.container);
         this.editor = editor;
+        /**
+        * @type Number
```
akoreman, line 137: I think this should be `@type {number | undefined}` (curly brackets around the type)? 

(see the examples here https://jsdoc.app/tags-type)

akoreman (approved):

PR #5711, opened 2025-01-06
docs: Correct return type of Search.find()

ferdnyc:
*Description of changes:*

The documentation comments for `Search.find()` claim it can return `Range` or `false` (changed from `Range` or `boolean` in 7e7f3298d), but neither is accurate. `firstRange` (the returned value) will always be either a `Range`, or `null` when the inner anonymous function returns `false`.

Pull Request Checklist:


akoreman:
Thanks for your contribution! 

If I read correctly, `this.$matchIterator(session, options)` could have a falsy return in which case this function will return `false` as well, so shouldn't the return type here be `Range | null | false`?

As a second note, could you also run `npm run update-types` to update the `*.d.ts` type declaration files.

ferdnyc:
@akoreman 

> If I read correctly, `this.$matchIterator(session, options)` could have a falsy return in which case this function will return `false` as well, so shouldn't the return type here be `Range | null | false`?

Eek, you're right. I missed the early return up top there. I'll fix that, thanks.

> As a second note, could you also run `npm run update-types` to update the `*.d.ts` type declaration files.

Shall do.

ferdnyc:
@akoreman 

Part of me now wonders, though, if the early return shouldn't just be,

```js
if (!iterator)
    return null;
```

...to avoid having to deal with two return types.

Although I suppose `find()` returning `false` rather than `null` (which happens iff `$matchIterator()` returns `false`, which in turn can only happen if `$assembleRegExp()` returns `false`) could be treated by the caller as an indication that something has gone wrong in the call, as opposed to a "no results" result.

akoreman (approved):

akoreman:
It's probably best to leave the actual return signature as-is to prevent potential breaking changes, thanks for correcting the type declaration 🚀

Issue #5716, opened 2025-01-11
Update selected text area at the user's refresh rate


thennothinghappened:
### Describe the feature

I know this is a nitpick that your average sensible person wouldn't care about whatsoever, _but_, when editing code on a high refresh-rate monitor (i.e., considerably greater than 60hz), a code editor (or any text editor that permits selection) is generally expected to match this smoothness when using the mouse to select text.

This is the case for the default browser behaviour, and in others such as Monaco, and feels a bit jarring when Ace then doesn't do this, and uses a fixed update interval.

### Use Case

This just feels a lot nicer! - and I'm sure that there's some others who use a high refresh rate monitor who'd appreciate it, but haven't been bothered enough by nearly enough it to ask.

### Proposed Solution

I did some very basic digging, and found the line responsible here: https://github.com/ajaxorg/ace/blob/ef08dfb6c5040f7493c2b30b7c19345e86c420c8/src/mouse/mouse_handler.js#L185

Here, `setInterval` is being used with a fixed update gap of `20` millis, which is the source of this slowness.

As a basic test I initially tried lowering this to `4`, which drastically improved the result on my machine.
However, this is not necessary on a machine that isn't running at a higher refresh rate. Thus, I then also tried replacing the use of `setInterval` with `requestAnimationFrame`, and replacing `timerId` with a boolean `timerShouldRun`, which solved the issue.

I noted that Ace seems to have some kind of built-in fallback options for `requestAnimationFrame` on older browsers which may not support it, so I'd imagine this utility would be used in a proper implementation.

I also noticed that this magic number `20` is being used in quite a few places throughout the codebase - whilst some appeared to be for spacing out logic, I wonder if some other parts of editor functionality could be made a bit smoother by applying this same trick.

Another thing to note here is that most monitors are 60hz, in a standard setting. This comes to `16.666...`ms per frame, which means that the use of `20` on most monitors is already going to be out of sync, thus resulting in a jittery look even on a standard monitor - I have a second monitor which is 60hz, and can confirm that by using `requestAnimationFrame`, it too feels _noticably_ smoother, at least to me.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### ACE version used

1.37.4

PR #5717, opened 2025-01-11
feat: update mouse selection at user's refresh rate

thennothinghappened:
*Issue #, if available:* #5716

*Description of changes:*
Rather than updating the mouse selection at a fixed interval of 20 milliseconds, I've changed it to use `event.nextFrame()`, which calls `requestAnimationFrame()`. This means that user selection is updated at their refresh rate, and thus results in a much smoother experience; this is especially noticeable on (but not limited to) high refresh-rate monitors.

I did note that there are many places within Ace where this fixed magic number of 20 is used, rather than hooking into the animation loop - these could also be made smoother by making similar adjustments.

Pull Request Checklist:

### DIFF
--- a/src/mouse/mouse_handler.js
+++ b/src/mouse/mouse_handler.js
@@ -129,7 +129,7 @@ class MouseHandler {
         renderer.$isMousePressed = true;
 
         var self = this;
-        var continueCapture = false;
+        var continueCapture = true;
 
         var onMouseMove = function(e) {
             if (!e) return;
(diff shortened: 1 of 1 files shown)


akoreman:
Thanks for creating the PR, I've been playing around with the change a bit and it feels nice and smooth. As [`event.nextFrame`](https://github.com/ajaxorg/ace/blob/master/src/lib/event.js#L364) falls-back to a normal timeout when browsers don't support `window.requestAnimationFrame` I also don't see any support issues.

akoreman (approved):

PR #5740, opened 2025-02-07
feat(popup): add support for skipping filtering with prefix by introducing skipFilter flag

nlujjawal:
*Description of changes:*
Add support for skipping filtering with prefix by introducing skipFilter flag which will all the popup item to not get filtered out incase if the popup item value is not matching the prefix.

Pull Request Checklist:
* [ yes] No backwards incompatible changes were made to Ace's public interface which is defined through the main typings file (`ace.d.ts`) and its references:

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -28,6 +28,7 @@ var preventParentScroll = require("./lib/scroll").preventParentScroll;
  * @property {string} [docText] - a plain text that would be displayed as an additional popup. If `docHTML` exists,
  * it would be used instead of `docText`.
  * @property {string} [completerId] - the identifier of the completer
+ * @property {boolean} [skipFilter] - a boolean value to decide if the popup item is going to skip the filtering process done using prefix text.
  * @property {import("../ace-internal").Ace.IRange} [range] - An object specifying the range of text to be replaced with the new completion value (experimental)
  * @property {any} [command] - A command to be executed after the completion is inserted (experimental)
  * @property {string} [snippet] - a text snippet that would be inserted when the completion is selected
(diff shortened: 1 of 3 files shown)


src/autocomplete.js
```diff
@@ -27,6 +27,7 @@ var preventParentScroll = require("./lib/scroll").preventParentScroll;
  * @property {string} [docHTML] - an HTML string that would be displayed as an additional popup
  * @property {string} [docText] - a plain text that would be displayed as an additional popup. If `docHTML` exists,
  * it would be used instead of `docText`.
+ * @property {Function} [clickCallback] - a callback function which can be associated with a popup item. If present text value won't be used after clicking the popup item.
```
akoreman, line 30: nitpick: looking at other [handlers](https://github.com/ajaxorg/ace/blob/fd61361d8307cc3d7307a16e7d96b228ed47d173/types/ace-modules.d.ts#L1607) we have for similar functionality, could it make sense to also call this something like `on<xyz>`, so maybe `onClick` or `onItemClick`

src/autocomplete.js
```diff
@@ -462,6 +463,10 @@ class Autocomplete {
             data = this.popup.getData(this.popup.getRow());
         if (!data)
             return false;
+        if (data.clickCallback) {
```
akoreman, line 466: We should probably add a test to verify the behavior 🙂

src/autocomplete.js
```diff
@@ -462,6 +463,10 @@ class Autocomplete {
...
+            data.clickCallback();
```
xyos, line 467: would it be useful to pass data to this function?

can we unit test this?

nlujjawal: We can already have the data from the completer which we set in popup data later. Yeah we can do unit test.

src/autocomplete.js
nlujjawal: ack

akoreman (approved):

xyos (approved):

nightwing (approved):

PR #5747, opened 2025-02-19
fix: gutter hover tooltip a11y improvements

babalaui:
*Issue #, if available:*

*Description of changes:*
- to make sure the screen reader reads the gutter hover tooltip, added tooltip role and ID, and referenced this ID with aria-describedby of gutter annotations
- made tooltip mouse hoverable
- tooltip will still close on any keypress, unless the CTRL/Command are also being pressed at the same time, in case the user wants to copy text from the tooltip
- improved tests by waiting for async tests to execute (some tests were failing on master too, if properly awaited)

Pull Request Checklist:


src/layer/gutter.js
```diff
@@ -524,6 +525,7 @@ class Gutter{
             annotationNode.setAttribute("aria-label", ariaLabel);
             annotationNode.setAttribute("tabindex", "-1");
             annotationNode.setAttribute("role", "button");
+            annotationNode.setAttribute("aria-describedby", GUTTER_HOVER_TOOLTIP_ID);
```
akoreman, line 528: Would  this set `aria-describedby` on all gutter annotations which are currently rendered, also on those which the open tooltip doesn't relate to? would that cause issues?

src/mouse/default_gutter_handler.js
```diff
@@ -71,20 +81,28 @@ function GutterHandler(mouseHandler) {
...
             }
         }
     }
 
-    function hideTooltip() {
+    function hideTooltip(e) {
+        // dont close tooltip in case the user wants to copy text from it (Ctrl/Meta + C)
+        if (e && e.type === "keydown" && (e.ctrlKey || e.metaKey))
+            return;
+        // dont close tooltip if mouse is on tooltip
```
akoreman, line 96: reading this comment it reads as if when I hover over a gutter annotation and the tooltip shows, pressing escape doesn't close it (as the mouse is still on the tooltip). But in the test we add that seems to be what happens, am I misreading this comment?

src/layer/gutter.js
babalaui: it would indeed set it on all annotations, I dont see it being an issue though since the tooltip is the main way of viewing those annotations, so technically all of them are described by it. do you think its better to only add this attribute to the annotation currently described by the tooltip?

src/mouse/default_gutter_handler.js
babalaui: the tooltip will close when pressing escape even if mouse is on the tooltip, so I'll change the comment to better reflect that

src/layer/gutter.js
akoreman: With aria-describedby always being set to the tooltip, when you keyboard navigate over the gutter annotations does it still properly read the aria-label of that annotation or does it read the tooltip (even when the tooltip isn't currently open)?

babalaui: ah indeed, it seems that besides the aria label it will also read the tooltip, resulting in cases where it would read the tooltip from the previous annotation. I'll see if I find an easy fix for this, if not I'll remove this and tackle it in another PR.

akoreman:
I think the first bullet-point of the PR description is no longer accurate, could you remove it to prevent future confusion 🙂

akoreman (approved):

PR #5771, opened 2025-03-14
Feat: Editing is disabled tooltip in readonly mode

mkslanc:
*Issue #, if available:*

*Description of changes:*

image

Pull Request Checklist:

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -2846,8 +2846,8 @@ config.defineOptions(Editor.prototype, "editor", {
                     if (!this.hoverTooltip) {
                         this.hoverTooltip = new HoverTooltip();
                     }
-                    var domNode = dom.createElement("div");
-                    domNode.innerText = "Editing is disabled";
+                    const domNode = dom.createElement("div");
+                    domNode.textContent = nls("editor.tooltip.disable-editing", "Editing is disabled");
                     if (!this.hoverTooltip.isOpen) {
                         this.hoverTooltip.showForRange(this, this.getSelectionRange(), domNode);
                     }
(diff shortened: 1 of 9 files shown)


src/editor.js
```diff
@@ -2836,9 +2838,32 @@ config.defineOptions(Editor.prototype, "editor", {
...
+        set: function(/**@type{boolean}*/readOnly) {
             this.textInput.setReadOnly(readOnly);
             this.$resetCursorStyle();
+            if (!this.readOnlyCallback) {
+                this.readOnlyCallback = () => {
+                    if (!this.hoverTooltip) {
+                        this.hoverTooltip = new HoverTooltip();
+                    }
+                    var domNode = dom.createElement("div");
+                    domNode.innerText = "Editing is disabled";
```
akoreman, line 2850: We should use the `nls` utility to allow integrators to translate this string into whatever locale they want.

mkslanc: Right, totally forgot about it, added nls message and covered by tests

marinsokol5 (approved):

akoreman (approved):

marinsokol5:
It's not related to this change, since we just reuse the HoverTooltip, but my VoiceOver on MacOS in Chrome doesn't read the tooltip content at all. Might work for other browser/OS/SR implementation though.

akoreman:
@mkslanc Thanks for fixing the failing build! It seems we introduced a merge conflict with this CR, could you resolve that conflict 🙂

mkslanc:
> @mkslanc Thanks for fixing the failing build! It seems we introduced a merge conflict with this CR, could you resolve that conflict 🙂

Done :) Hope it helped

PR #5781, opened 2025-03-31
feature(gutter): add ability to hide fold icon for a specific row

nlujjawal:
*Description of changes:*
Adding ability to hide the fold icon for a specific row.

### DIFF
--- a/types/ace-modules.d.ts
+++ b/types/ace-modules.d.ts
@@ -422,6 +422,11 @@ declare module "ace-code/src/layer/gutter" {
         getShowLineNumbers(): boolean;
         setShowFoldWidgets(show?: boolean): void;
         getShowFoldWidgets(): boolean;
+        /**
+         * Hides the fold widget/icon from a specific row in the gutter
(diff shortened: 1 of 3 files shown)


types/ace-modules.d.ts
```diff
@@ -3948,6 +3953,11 @@ declare module "ace-code/src/edit_session" {
          * @param {String} className The class to add
          **/
         addGutterDecoration(row: number, className: string): void;
+        /**
+         * Hides the fold widget/icon from a specific row in the gutter
```
akoreman, line 3957: We're adding an API to hide gutter icons on a row, but how can users unhide those hidden icons?

akoreman:
More general notes:
- can we add a test verifying that the fold icons are indeed hidden and no longer interactive?
- what's the usecase we see for having a public API method for hiding individual folding icons?

nlujjawal:
PR is in construction mode.

nlujjawal:
PR is in construction mode.

PR #5787, opened 2025-04-03
feature(gutter): add experimental ability to replace fold icon with custom for a specific row

nlujjawal:
*Issue #, if available:*

*Description of changes:*
Adding Experimental capability in gutter to replace the fold icon with the custom icon.
Custom widget Icon can be added which will hides the fold Icon and show custom widget icon.
And custom widget Icon can be removed which will remove the custom widget and displays the fold icon.

### DIFF
--- a/src/edit_session.js
+++ b/src/edit_session.js
@@ -590,20 +590,21 @@ class EditSession {
      * @param {number} row The row number for which to hide the custom icon
      */
     removeGutterCustomWidget(row) {
-        this.$editor.renderer.$gutterLayer.removeCustomWidget(row);
+        this.$editor.renderer.$gutterLayer.$removeCustomWidget(row);
     }
 
     /**
      * Replaces the fold widget if present with the custom icon from a specific row in the gutter
      * @param {number} row - The row number where the widget will be displayed
-     * @param {Object} options - Configuration options for the widget
-     * @param {string} options.className - CSS class name for styling the widget
-     * @param {string} options.label - Text label to display in the widget
-     *  @param {string} options.title - Tooltip text for the widget
+     * @param {Object} attributes - Configuration attributes for the widget
+     * @param {string} attributes.className - CSS class name for styling the widget
+     * @param {string} attributes.label - Text label to display in the widget
+     * @param {string} attributes.title - Tooltip text for the widget
+     * @param {Object} attributes.callbacks - Event callback functions for the widget e.g onClick; 
      * @returns {void}
     */
     addGutterCustomWidget(row,attributes) {
-        this.$editor.renderer.$gutterLayer.addCustomWidget(row,attributes);
+        this.$editor.renderer.$gutterLayer.$addCustomWidget(row,attributes);
     }
 
     /**
--- a/src/edit_session.js
+++ b/src/edit_session.js
@@ -594,7 +594,9 @@ class EditSession {
      * @experimental
      */
     removeGutterCustomWidget(row) {
-        this.$editor.renderer.$gutterLayer.$removeCustomWidget(row);
+        if(this.$editor) {
+            this.$editor.renderer.$gutterLayer.$removeCustomWidget(row);
+        }
     }
 
     /**
@@ -609,7 +611,9 @@ class EditSession {
      * @experimental
     */
     addGutterCustomWidget(row,attributes) {
-        this.$editor.renderer.$gutterLayer.$addCustomWidget(row,attributes);
+        if(this.$editor) {
+            this.$editor.renderer.$gutterLayer.$addCustomWidget(row,attributes);
+        }
     }
 
     /**
--- a/src/edit_session.js
+++ b/src/edit_session.js
@@ -586,6 +586,27 @@ class EditSession {
     }
 
     /**
+     * Replaces the custom icon with the fold widget if present from a specific row in the gutter
+     * @param {number} row The row number for which to hide the custom icon
+     */
+    removeGutterCustomWidget(row) {
+        this.$editor.renderer.$gutterLayer.removeCustomWidget(row);
+    }
+
+    /**
+     * Replaces the fold widget if present with the custom icon from a specific row in the gutter
+     * @param {number} row - The row number where the widget will be displayed
+     * @param {Object} options - Configuration options for the widget
+     * @param {string} options.className - CSS class name for styling the widget
+     * @param {string} options.label - Text label to display in the widget
+     *  @param {string} options.title - Tooltip text for the widget
+     * @returns {void}
+    */
+    addGutterCustomWidget(row,attributes) {
+        this.$editor.renderer.$gutterLayer.addCustomWidget(row,attributes);
+    }
+
+    /**
      * Removes `className` from the `row`.
      * @param {Number} row The row number
      * @param {String} className The class to add
--- a/src/edit_session.js
+++ b/src/edit_session.js
@@ -57,6 +57,9 @@ class EditSession {
             return this.join("\n");
         };
 
+        // @experimental
+        this.$gutterCustomWidgets = {};
+
         // Set default background tokenizer with Text mode until editor session mode is set
         this.bgTokenizer = new BackgroundTokenizer((new TextMode()).getTokenizer(), this);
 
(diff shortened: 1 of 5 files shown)


src/layer/gutter.js
```diff
@@ -588,6 +593,89 @@ class Gutter{
...
+    * @returns {void}
+    */
+    showCustomWidget(row, {className, label, title}) {
+        const cells = this.$lines.cells;
+        const cell = cells[row];
+        if (cell && cell.element) {
+            var customWidget = cell.element.querySelector(".ace_custom-widget");
+                this.hideFoldWidget(row);
+                // if customWidget already exists show it.
+            if (customWidget) {
```
akoreman, line 643: Would this mean that a user is unable to change what a widget looks like at a given row? e.g. if they at some point show widget A at row 8, would they be unable to show widget B at that some row at a later time? I do think that's an use case we need to support

nlujjawal: Yeah probably for future features scope may be like diff view we might need it. Good catch I will modify it.

src/layer/gutter.js
```diff
@@ -588,6 +593,80 @@ class Gutter{
...
+    * Remove a custom widget for a specific row
+    * @param {number} row - The row number where the widget will be removed
+    * @returns {void}
+    */
+    removeCustomWidget(row){
+        const cells = this.$lines.cells;
+        const cell = cells[row];
+        if (cell && cell.element) {
+            const customWidget = cell.element.querySelector(".ace_custom-widget");
+            cell.element.removeChild(customWidget);
```
akoreman, line 666: What happens if you call `removeCustomWidget` for a row for which there is no widget?

nlujjawal: ACK added a check for that.

akoreman (approved):

src/layer/gutter.js
```diff
@@ -331,6 +331,7 @@ class Gutter{
         
         var textNode = element.childNodes[0];
         var foldWidget = element.childNodes[1];
+        var customWidget = element.childNodes[3];
```
xyos, line 334: nit: can we move this after next line? so it's on the same index order

src/layer/gutter.js
```diff
@@ -588,6 +593,82 @@ class Gutter{
...
+    showFoldWidget(row) {
+        const cells = this.$lines.cells;
+        const cell = cells[row];
+        if (cell && cell.element) {
+            const foldWidget = cell.element.childNodes[1];
+            if (foldWidget) {
+                dom.setStyle(foldWidget.style, "display", "inline-block");
+            }
+        }
+    }
```
xyos, line 625: should this 2 be a `toggleFoldWidget(row)`?

src/edit_session.js
```diff
@@ -585,6 +585,27 @@ class EditSession {
...
+        this.$editor.renderer.$gutterLayer.removeCustomWidget(row);
+    }
+
+    /**
+     * Replaces the fold widget if present with the custom icon from a specific row in the gutter
+     * @param {number} row - The row number where the widget will be displayed
+     * @param {Object} options - Configuration options for the widget
+     * @param {string} options.className - CSS class name for styling the widget
+     * @param {string} options.label - Text label to display in the widget
+     *  @param {string} options.title - Tooltip text for the widget
```
xyos, line 602:
```suggestion
     * @param {Object} attributes - Configuration options for the widget
     * @param {string} attributes.className - CSS class name for styling the widget
     * @param {string} attributes.label - Text label to display in the widget
     * @param {string} attributes.title - Tooltip text for the widget
```

types/ace-modules.d.ts
```diff
@@ -422,6 +422,34 @@ declare module "ace-code/src/layer/gutter" {
...
+        /**
+         * Hides the fold widget/icon from a specific row in the gutter
+         * @param {number} row The row number from which to hide the fold icon
+         */
+        hideFoldWidget(row: number): void;
+        /**
+         * Shows the fold widget/icon from a specific row in the gutter
+         * @param {number} row The row number from which to show the fold icon
+         */
+        showFoldWidget(row: number): void;
```
akoreman, line 434: Should we mark these methods as internal so that they don't show up in our public `.d.ts` files? Or do we see a clear use case for customers using these method?

src/layer/gutter.js
```diff
@@ -418,6 +419,10 @@ class Gutter{
             // Set a11y properties.
             foldWidget.setAttribute("role", "button");
             foldWidget.setAttribute("tabindex", "-1");
+
+            if(customWidget && window.getComputedStyle(customWidget).getPropertyValue('display') !== 'none'){
```
nightwing, line 423: `getComputedStyle` is too slow to use it in render call, also it can cause relayout, so we should find another way to do this

src/layer/gutter.js
```diff
@@ -588,6 +593,82 @@ class Gutter{
...
         return this.$showFoldWidgets;
     }
+    
+    /**
+     * Hides the fold widget/icon from a specific row in the gutter
+     * @param {number} row The row number from which to hide the fold icon
+     */
+    hideFoldWidget(row) {
+        const cells = this.$lines.cells;
+        const cell = cells[row];
```
nightwing, line 603: cells has only the visible lines, so row is not the actual document row

nightwing (changes requested):

src/layer/gutter.js
nlujjawal: added getguttercell method to over come this issue.

src/layer/gutter.js
nlujjawal: it will lead a lot of if else and can lead to less readability

src/layer/gutter.js
```diff
@@ -588,6 +599,113 @@ class Gutter{
...
+    
+    /**
+    * Retrieves the gutter cell element at the specified cursor row position.
+    * @param {number} row - The row number in the editor where the gutter cell is located starts from 0
+    * @returns {HTMLElement|null} The gutter cell element at the specified row, or null if not found
+    */
+    $getGutterCell(row){
+        // contains only visible rows
+        const cells = this.$lines.cells;
+        const visibileRow= this.session.documentToScreenRow(row,0);
```
nightwing, line 641: documentToScreenRow on large documents can be much slower than searching row in the cells array

src/layer/gutter.js
xyos: only lines 607 and 622 are uncommon and it would be a simple ternary, do you still think it would be less readable?

akoreman: if we merge them, I would maybe call it something like `toggleFoldWidgetVisibility` to make sure it's clear we are toggling the visibility of the widget and are not folding/unfolding the code at that widget.

src/layer/gutter.js
nlujjawal: Searching the cells doesn't work properly in case when the line is folded and you move the cursor to the right of the fold symbol in the code. It doesn't work because at that time cursor is on a hidden row.

I tried creating a large document of 50K lines current code seems to be working fine without any lag.

src/layer/gutter.js
nlujjawal: There is 2 lines(one is if condition) of difference now in both the functions and if we add toggle logic,
There will be 2 extra nested `if's` with two more of the `if's` which will always exists which I think won't be that much readable.

akoreman (approved):

xyos (approved):

src/layer/gutter.js
nlujjawal: ack removed that

src/edit_session.js
```diff
@@ -585,6 +588,30 @@ class EditSession {
         this._signal("changeBreakpoint", {});
     }
 
+    /**
+     * Replaces the custom icon with the fold widget if present from a specific row in the gutter
+     * @param {number} row The row number for which to hide the custom icon
+     * @experimental
+     */
+    removeGutterCustomWidget(row) {
+        this.$editor.renderer.$gutterLayer.$removeCustomWidget(row);
```
nightwing, line 597: `$editor` can be null

nlujjawal: acknowledged that.

nightwing (approved):
i am not convinced that this is a good api addition, but approving as experimental

PR #5793, opened 2025-04-09
unfold command should not add folds

nightwing:


xyos (approved):

src/edit_session/folding.js
```diff
@@ -649,6 +649,13 @@ function Folding() {
             if (fold) {
                 this.expandFold(fold);
                 return;
+            } else if (tryToUnfold) {
+                range.start.column = 0;
+                range.end.column = Infinity;
+                var folds = this.getFoldsInRange(range);
+                if (tryToUnfold && folds.length)
```
marinsokol5, line 656: tryToUnfold is already checked above to be truthy

marinsokol5 (approved):

src/edit_session/folding.js
nightwing: thanks, fixed.

PR #5796, opened 2025-04-14
feat(gutter): added keyboard handling for experimental feature custom widgets

nlujjawal:
*Description of changes:*
Added keyboard handling for experimental feature custom widgets which required changes for the code which handles when gutter is active and user presses enter, up, down, space etc.

### DIFF
--- a/src/keyboard/gutter_handler.js
+++ b/src/keyboard/gutter_handler.js
@@ -54,24 +54,21 @@ class GutterKeyboardHandler {
                 /** @this {GutterKeyboardHandler} */
                 function () {
                     var index = this.$rowToRowIndex(this.gutterLayer.$cursorCell.row);
-                    var nearestFoldWidgetIndex = this.$findNearestFoldWidget(index);
-                    var nearestCustomWidgetIndex = this.$findNearestCustomWidget(index);
-                    var nearestFoldLaneIndex = this.$findClosestNumber(nearestFoldWidgetIndex, nearestCustomWidgetIndex, index);
+                    var nearestFoldLaneWidgetIndex = this.$findNearestFoldLaneWidget(index);
                     var nearestAnnotationIndex = this.$findNearestAnnotation(index);
 
-                    if (nearestFoldLaneIndex === null && nearestAnnotationIndex === null) return;
+                    if (nearestFoldLaneWidgetIndex === null && nearestAnnotationIndex === null) return;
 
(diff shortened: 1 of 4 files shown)


src/keyboard/gutter_handler.js
```diff
@@ -205,6 +206,7 @@ class GutterKeyboardHandler {
             switch (this.activeLane){
                 case "fold":
                     this.$blurFoldWidget(this.activeRowIndex);
```
akoreman, line 208: Does `$blurFoldWidget` here expect there always to be a foldwidget at the provided row or will it handle the possibility of no fold widget being there okay?

src/keyboard/gutter_handler.js
```diff
@@ -265,6 +279,28 @@ class GutterKeyboardHandler {
         return null;
     }
 
+    // Given an index, find the nearest index with a customwidget
+    $findNearestCustomWidget(index) {
```
akoreman, line 283: bit of a nitpick given that these are all private methods, but given that fold widgets and custom widgets are mutually exclusive (a row can only either have a fold widget or a custom widget), would it make sense to just have a single method `findNearestFoldLaneWidget` (or something along those lines) to find whatever widget is closest in the fold lane (same for all other methods involving fold and custom widgets)

nlujjawal: Ack.
Merging nearestCustomWidget and nearestFoldWidget makes sense, for other methods it doesn't make much sense as it will increase the lines of code requried.

akoreman (approved):

src/keyboard/gutter_handler.js
```diff
@@ -237,22 +242,37 @@ class GutterKeyboardHandler {
...
         // If fold widget exists at index, return index.
         if (this.$isFoldWidgetVisible(index))
             return index;
 
-        // else, find the nearest index with fold widget within viewport.
+        // else, find the nearest index with widget within viewport.
         var i = 0;
         while (index - i > 0 || index + i < this.lines.getLength() - 1){
             i++;
+            if (index - i >= 0 && this.editor.session.$customWidgets[this.$rowIndexToRow(index - i)])
```
akoreman, line 271: Why aren't we using `isCustomWidgetVisible` here?

nlujjawal: Yeah updated, both things can be same its equally same but for consistency updated that.

PR #5798, opened 2025-04-21
import diff viewer

nightwing:
https://raw.githack.com/ajaxorg/ace/78a7d8586b184e300301bcae29b0c88cf5465af4/demo/diff/index.html


nlujjawal:
Awesome PR diff view looks really nice. I was playing around and found some bugs

1.  alignment of fold icons is not consistent for a normal line vs a removed line.
Screenshot 2025-04-22 at 15 54 17
2. There is empty space in the bottom when the code is folded in the removed lines 
Screenshot 2025-04-22 at 15 54 33
3. There is a very little green color bar in the removed lines fold
Screenshot 2025-04-22 at 15 55 24

marinsokol5:
Does it make sense to disable folding all together when viewing the diff? 
Cause when added lines are inside of the current content, like in the one below, folding it would hide the added line. It's kind of unclear what would be the best UX for it, so might indeed be best to not allow it at all.
Screenshot 2025-04-23 at 10 43 48

We can also check how Monaco inline diff acts when folding, but I would lean on the side of disabling.

xyos:
Great work 🥳  I found another bug with the collapse unchanged lines option and it's related to unfolding with a similar behavior to the one @akoreman reported:

image

marinsokol5:
Is dual numbering in the gutter intentional?

nightwing:
I have fixed most of the issues with folds except the case when fold starts outside of changed range and ends inside one, perhaps we should indeed disable folding in that case.
I do not want to disable folding altogether, since it is useful for collapsing unchanged lines

For the accept/reject feauture, i think the easiest solution will be to modify align method to include widgets for these too.

I've also added ability to select text inside the green zones, but not sure if we should limit selection to one green zone, or to allow to select as if we are in a diffview with new code editable. In principle using the approach we could make both old and new editable in inline view.

akoreman (approved):

PR #5800, opened 2025-04-23
Improve JSDoc annotations for code lens

xyos:
*Issue #, if available:*

*Description of changes:*

This PR adds JSDocs to provide typescript support for the codelens functionality

Pull Request Checklist:

### DIFF
--- a/ace.d.ts
+++ b/ace.d.ts
@@ -222,7 +222,7 @@ declare module "ace-code" {
             fullWidth?: boolean;
             screenWidth?: number;
             rowsAbove?: number;
-            lenses?: any[];
+            lenses?: CodeLenseCommand[];
         }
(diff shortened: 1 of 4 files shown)


ace.d.ts
```diff
@@ -875,14 +875,56 @@ declare module "ace-code" {
...
-            start: Point;
-            command: any;
+            /**
+             * Starting position where the code lens should be displayed
+             */
+            start: Point,
+            /**
+             * Command to execute when the code lens is activated
+             */
+            command: CodeLenseCommand
```
akoreman, line 919: changing the type here from `any` to `CodeLenseCommand` could technically be seen as a breaking change, do you think there is a likely scenario where consumers would be using any other type than `CodeLenseCommand` here?

akoreman (approved):

PR #5823, opened 2025-06-03
Add support for the Clue language

Markos-Th09:
*Issue #, if available:*

*Description of changes:*
Add support for the [Clue](https://github.com/ClueLang/Clue) programming language. The syntax highlighter was generated from [this grammar](https://github.com/ClueLang/Clue-for-VSCode/blob/main/syntaxes/clue.json).

Pull Request Checklist:


xyos:
Hey @Markos-Th09, thanks for your contribution, can you fix the rules for comment?
```
broken blockComment in clue {
  tokens: [
    { type: 'punctuation.definition.comment.begin.clue', value: '/*' },
    { type: 'comment.block.clue', value: '/* ' },
    { type: 'punctuation.definition.comment.end.clue', value: '*/' },
    { type: 'keyword.operator.clue', value: '*/' }
  ],
  state: 'start'
}
```

Markos-Th09:
@xyos I am not exactly sure what I need to change in my grammar for it to not be broken. what is this testing exactly?

xyos:
Hi @Markos-Th09, the [test is failing](https://github.com/ajaxorg/ace/blob/master/src/mode/_test/highlight_rules_test.js#L107-L107) because your [grammar's operator rule](https://github.com/ajaxorg/ace/pull/5823/files#diff-ec742f2b771b6dadba626a7d56e0d129ce53eeabebd4c664b6956d4c0f75d693R155) matches * and / as operators in the main state, so when a block comment ends (with */), those characters can be misclassified as operators instead of the end of a comment. This causes the tokenizer to break block comment highlighting, especially for cases like nested or chained comments, which is what the test is catching. The issue isn't with your comment rules themselves, but with the interaction between the operator and comment rules in your tokenizer states so `/* /* */*/ */` would break the tokenizer in this case.

Markos-Th09:
@xyos This is correct behavior. I mistakenly allowed nesting, when it is not actually supported. I have now disabled it and the tests for clue should now be passing.

Markos-Th09:
> One of the public type files has been updated, plase make sure there are no backwards incompatible changes done in the PR.

Just including the new type definitions for the Clue mode and highlighter

Markos-Th09:
>   72:21  error  Parsing error: Invalid regular expression: /(?<![\w\d.])0[xX][0-9A-Fa-f]+(?![pPeE.0-9])/: Invalid group

Are negative lookbehinds not allowed for compatibility?
What mechanism of the tokenizer could replace them for those that can’t be rewritten using word boundaries easily?

xyos:
@nightwing I notice this feature relies on lookbehind assertions, which have had [broad browser support since March 2023](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion#browser_compatibility) when Safari added support as the last major browser.

Given that lookbehinds are now widely supported, should we reconsider whether this compatibility test is still necessary?

@Markos-Th09 I think you can use  word boundary `\b` instead

Markos-Th09:
@xyos Word boundary suffices for most of them, except one of the regexes which I rewrote using FSA rules.

xyos (approved):
Thanks @Markos-Th09 let me know if this is enough for now 😄

Markos-Th09:
@xyos It should be good for now. I'll make a follow up PR if anything else is needed.

PR #5825, opened 2025-06-05
Add `createDiffView` wrapper

mkslanc:
*Issue #, if available:*

*Description of changes:*
Add `createDiffView` wrapper to improve API and allow users to use extension in ace-builds 

Pull Request Checklist:


nightwing (approved):

ace.d.ts
```diff
@@ -40,8 +40,8 @@ declare module "ace-code" {
         type GutterKeyboardEvent = import("ace-code/src/keyboard/gutter_handler").GutterKeyboardEvent;
         type HoverTooltip = import("ace-code/src/tooltip").HoverTooltip;
         type Tooltip = import("ace-code/src/tooltip").Tooltip;
-        type PopupManager = import("ace-code/src/tooltip").PopupManager;
```
akoreman, line 43: Is it intentional to remove `PopupManager` from our types here?

src/ext/diff.js
```diff
@@ -0,0 +1,31 @@
+var InlineDiffView = require("./diff/inline_diff_view").InlineDiffView;
+var DiffView = require("./diff/diff_view").DiffView;
```
marinsokol5, line 2: We should call this one the split diff view :)

ace.d.ts
```diff
@@ -1031,6 +1031,52 @@ declare module "ace-code" {
             setLabel?: boolean;
             inline?: boolean;
         }
+        /**
+         * Interface representing a model for handling differences between two views or states.
+         */
+        export interface DiffModel {
```
marinsokol5, line 1037: do we need to add these types into the main ace.d.ts file? Can it be enough to just have these in ace-ext?

src/ext/diff.js
mkslanc: Should I make an alias for that purpose? Or just rename the class?

ace.d.ts
mkslanc: I guess we could, but in terms of reusing better to do it through some `.d.ts` file (and ace-ext.d.ts is auto-generated).  The only way to do this with js codebase (which I know) would be making jsdoc with `@typedef` tag. Also previously all interfaces were centralized in `ace.d.ts` under Ace namespace, not sure if that approach changed. Just poke me, if you still want it to be `@typedef` as [PromptOptions](https://github.com/ajaxorg/ace/blob/master/src/ext/prompt.js#L21) for example

marinsokol5 (approved):

akoreman (approved):

PR #5835, opened 2025-06-27
Feat: Implement text markers

mkslanc:
*Issue #, if available:*

*Description of changes:* 
This PR introduces a robust text markers implementation, enabling customized text highlighting. The feature includes a clean API, efficient rendering, and support for multiline selections.

*Key Features*
- Added text markers API for editor sessions
- Implemented DOM manipulation for text highlighting
- Added scheduling system to automatically remove outdated markers
- Enhanced search box functionality with selection highlighting (for kitchen-sink as a demo)

Pull Request Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [1.43.1](https://github.com/ajaxorg/ace/compare/v1.43.0...v1.43.1) (2025-07-02)
+
+
+### Bug Fixes
+
+* **a11y:** aria-describedby to gutter elements ([#5838](https://github.com/ajaxorg/ace/issues/5838)) ([bf27c15](https://github.com/ajaxorg/ace/commit/bf27c151685ac486a991a61d67fa04adf8891c76))
+
+### [1.42.1](https://github.com/ajaxorg/ace/compare/v1.42.0...v1.42.1) (2025-06-20)
+
+
+### Features
+
+* Add scroll diff decorators ([#5807](https://github.com/ajaxorg/ace/issues/5807)) ([b45e94c](https://github.com/ajaxorg/ace/commit/b45e94cb92d757b5f09ee0966a764ed0d245ed70))
+
 ## [1.42.0](https://github.com/ajaxorg/ace/compare/v1.41.0...v1.42.0) (2025-06-03)
 
 
--- a/ace-internal.d.ts
+++ b/ace-internal.d.ts
@@ -1439,9 +1439,14 @@ declare module "./src/editor" {
 }
 
 declare module "./src/edit_session" {
+    type TextMarkers = typeof import("./src/layer/text_markers").editSessionTextMarkerMixin & {
+        $textMarkers: any[];
+        $textMarkerId: number;
+    };
+
     export interface EditSession extends Ace.EventEmitter<Ace.EditSessionEvents>,
         Ace.OptionsProvider<Ace.EditSessionOptions>,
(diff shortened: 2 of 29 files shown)


marinsokol5:
Hey, great job on the initial draft! 🚀

Is there a nice way to trigger it today in the kitchen-sink or somewhere, or would one have to invoke the `addTextMarker` manually?
I am very curious to see how the UI would look like with this partial styling of certain tokens, where "mySt" is colored green, "rin" colored as text, and then "g" as green again.

Did you give any thought on how this will be introduced? Would something like a general editor option "disableSelectionHighlight" make sense, or is too specific, and we would maybe apply these text markers by default when there is a selection and let consumers handle it through CSS?

mkslanc:
@marinsokol5 Hey! I have [small kitchen sink demo](https://raw.githack.com/mkslanc/ace/color-change-selection-example/kitchen-sink.html) (CTRL-F would trigger colorization), that should show how it works with different tokens. 

This PR simply provides the low-level hooks for all kinds of highlighting scenarios—custom searches, high-contrast themes, ad-hoc token styling, etc.—and lets us iterate on the UX/API in follow-up discussions.

[screen-capture (96).webm](https://github.com/user-attachments/assets/f686d884-4c4f-4a39-9777-6f7529af7838)

marinsokol5:
Demo looks great, UI better than expected, especially with this partial token styling! Let's continue working on it and merge!

mkslanc:
@marinsokol5 hey! I've finished the PR and embedded the demo into the kitchen sink. Quick question - do we need a separate option in the kitchen sink panel to enable/disable it, or is it okay to have it permanently turned on?

akoreman:
> Quick question - do we need a separate option in the kitchen sink panel to enable/disable it, or is it okay to have it permanently turned on?

My two cents, it might be confusing for users to see this styling in the kitchen sink as they might think it's out-of-the-box ace styling (in contrast to something they need to add themselves if they want it)

mkslanc:
> > Quick question - do we need a separate option in the kitchen sink panel to enable/disable it, or is it okay to have it permanently turned on?
> 
> My two cents, it might be confusing for users to see this styling in the kitchen sink as they might think it's out-of-the-box ace styling (in contrast to something they need to add themselves if they want it)

I agree, but not sure, what would be the good way to bring it in demo

akoreman:
> I agree, but not sure, what would be the good way to bring it in demo

Yeah I'm also not sure, could it make sense to not have it in the kitchen sink for now so that we can merge the PR and then add a demo for it later?

mkslanc:
> > I agree, but not sure, what would be the good way to bring it in demo
> 
> Yeah I'm also not sure, could it make sense to not have it in the kitchen sink for now so that we can merge the PR and then add a demo for it later?

Removed it for now; btw: I almost finished working on PR based on that one, which would provide extension to show whitespaces in selection.

[screen-capture (99).webm](https://github.com/user-attachments/assets/0151cff4-fc3d-4c64-8111-7ad32eff9ec3)

marinsokol5 (approved):

marinsokol5:
Oh I liked the demo in the kitchen sink! 
I do agree that red was a bit too much 😄 . I would be okay to set it as black text color, so we make the search selection in the kitchen sink accessible while demoing the feature at the same time.

Screenshot 2025-07-14 at 16 10 24

mkslanc:
> Oh I liked the demo in the kitchen sink! I do agree that red was a bit too much 😄 . I would be okay to set it as black text color, so we make the search selection in the kitchen sink accessible while demoing the feature at the same time.
> 
> Screenshot 2025-07-14 at 16 10 24

It would still take some extra effort - at least to detect light/dark themes and switch between black and white color. :)

PR #5838, opened 2025-06-30
fix(a11y): update aria-describedby gutter annotation node on gutter popup show

mkslanc:
*Issue #, if available:*

*Description of changes:*

Pull Request Checklist:

### DIFF
--- a/src/mouse/default_gutter_handler.js
+++ b/src/mouse/default_gutter_handler.js
@@ -142,11 +142,13 @@ exports.GutterHandler = GutterHandler;
 class GutterTooltip extends Tooltip {
     constructor(editor, isHover = false) {
         super(editor.container);
+        this.id = "gt" + (++GutterTooltip.$uid);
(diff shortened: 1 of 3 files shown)


akoreman (approved):

src/mouse/default_gutter_handler.js
```diff
@@ -142,11 +142,13 @@ exports.GutterHandler = GutterHandler;
 class GutterTooltip extends Tooltip {
     constructor(editor, isHover = false) {
         super(editor.container);
+        this.id = "gt" + (++GutterTooltip.$uid);
```
marinsokol5, line 145: This seems a bit generic tbh, but i see it's what we used elsewhere too. Cause Ace can be embedded into any page, and id of "gt-0" or "gt-1" just doesn't feel scoped enough.

marinsokol5 (approved):

Issue #5834, opened 2025-06-22
ace/autocomplete: Autocompletion does not trigger when typing \ in latex mode


JeanCASPAR:
### Describe the bug

The autocompletion doesn't start when typing "\\", only when typing the letter next to it, which is troublesome when autocompleting LaTeX command because the "\\" will be written twice when accepting the suggestion.

### Expected Behavior

I expect the autocompletion to start running when typing "\\".

### Current Behavior

Nothing happens when typing "\\".

### Reproduction Steps

[...]

PR #5840, opened 2025-07-05
Add snippet support for LaTeX and custom macro completer.

JeanCASPAR:
*Issue #, if available:*
Fix [#5834](https://github.com/ajaxorg/ace/issues/5834).

*Description of changes:*
The snippets are copied from the snippets for TeX. The default keywordCompleter is unused in LaTeX mode because the highlight rules do not give a list of keywords, but color every LaTeX macro, so the completer cannot use the rules to provide completion. This macro completer is inspired from the text completer, but complete only macros (the current text completer do not consider \ to be a part of a word).

The interface of Completer is changed so that the keywordCompleter can forward the triggerCharacters and identifiersRegexps of `session.$mode.completer`. I don't know if this change is backward compatible, but I do not see any other way to achieve this goal. Maybe we can provide a better way for a mode to propose a completer than using keywordCompleter ?

Pull Request Checklist:

### DIFF
--- a/ace-internal.d.ts
+++ b/ace-internal.d.ts
@@ -976,7 +976,7 @@ export namespace Ace {
 
     interface Completer {
         /** Regular expressions defining valid identifier characters for completion triggers */
-        identifierRegexps?: Array<RegExp>,
+        identifierRegexps?: Array<RegExp> | ((editor: Editor) => Array<RegExp>),
 
         /** Main completion method that provides suggestions for the given context */
         getCompletions(editor: Editor,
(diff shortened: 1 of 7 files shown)


JeanCASPAR:
Also, the current mechanism of identifierRegexps is quite unsatisfactory : as we pass a macro that test every character separately, is it impossible for the function `getCompletionPrefix` in `src/autocomplete/util.js` to return only `\macro2` if I type `\macro1\macro2`, so I cannot get autocompletion when typing `\macro2`.

PR #5847, opened 2025-07-23
fix: the text completer didn't export its id

JeanCASPAR:
*Issue #, if available:*

*Description of changes:*
The text completer in text `src/autocomplete/text_completer.js` now has an ID, which makes things easier to debug. 

Pull Request Checklist:


nightwing:
Thank you

PR #5869, opened 2025-09-05
fix gutter custom widget modifying wrong rows

nightwing:


src/layer/gutter.js
```diff
@@ -638,11 +638,25 @@ class Gutter{
...
-        const cells = this.$lines.cells;
-        const visibileRow= this.session.documentToScreenRow(row,0);
-        // subtracting the first visible screen row index and folded rows from the row number.
-        return cells[row - this.config.firstRowScreen - (row-visibileRow)];
+        var cells = this.$lines.cells;
+        var min = 0;
+        var max = cells.length - 1;
+        
+        if (row < cells[0].row || row > cells[max].row)
+            return;
```
akoreman, line 646: Looking at the TSdoc above, should this be `return null`?

akoreman (approved):

PR #5882, opened 2025-10-21
ci: scope down GitHub Token permissions

AdnaneKhan:
## Scope Down GitHub Token Permissions

This PR updates GitHub Actions workflows to use minimal required permissions instead of the default elevated permissions.

### Why This Matters

Following the principle of least privilege, workflows should only have the specific permissions they need to function.

### Changes

This PR adds explicit `permissions:` blocks to workflows that currently rely on default permissions, scoping them down to only what's required for their operations.

Please review the changes to ensure the specified permissions match your workflow requirements.


akoreman:
Apologies for the delay in the review, I've merged the PR. Thanks for your contribution!

Issue #5813, opened 2025-05-21
Emojis consisting of surrogate pairs are displayed incorrectly


iamqwertyloal:
### Describe the bug

When I insert emojis containing gender or skin tone modifiers, the editor splits them into multiple characters

### Expected Behavior

Emojis are displayed correctly

### Current Behavior

After inserting an emoji, instead of one character, several are displayed, split by `<span class="ace_invisible ace_invisible_space ace_invalid">·</span>` elements

### Reproduction Steps

Insert these emojis into the editor 🚵‍♂️🚴‍♀️ and you will get this: 

![Image](https://github.com/user-attachments/assets/053a1aff-352c-4fef-8dce-1f14724f91c2)

### Ace Version / Browser / OS / Keyboard layout

ace-builds 1.32.5 / Chrome 136.0.7103.59 / Ubuntu 22.04

whazor:
I think this a duplicate of #4142

iamqwertyloal:
> We can remove `\u200d` from the list of characters in https://github.com/ajaxorg/ace/blob/v1.40.1/src/layer/text.js#L353 to not show that as an invalid hidden character and show it as part of emoji

That would be great! Could you create a PR with the necessary changes?

Issue #5431, opened 2023-12-18
Full Unicode support incomplete


stefanocoretta:
### Describe the bug

This is an issue that has been brought up several times through the years but it still not fixed (see for example #1531).

The editor does not handle certain Unicode characters correctly, for example the `Combining Diacritical Marks`. It seems the editor treats these characters as full width characters and the text cursor moves further away from the text.

I described the problem in details here, including pictures: https://github.com/rstudio/rstudio/issues/12680#issue-1575708046 RStudio uses the Ace editor.

### Expected Behavior

Correct handling of Unicode characters.

### Current Behavior

See description.

### Reproduction Steps

See description.

### Ace Version / Browser / OS / Keyboard layout

Ace Kitchen Sink, Safari, macOS; also embedded editor in RStudio latest

whazor:
Hi, I verified that the characters `ã̤n̺ ɾ̺ub̰ĩ̈`, which are monospaced characters get an incorrect double space.

I wonder if this issue would be different from https://github.com/ajaxorg/ace/issues/1153

stefanocoretta:
Yes, it's related to that. :) Do let me know if I can help with testing or the like. I'm happy to provide you with more details about characters/Unicode codes etc if it can help.

bradrn:
Any progress on this issue? It’s severely affecting my application https://bradrn.com/brassica, for which support for combining diacritics is very important.

ewnjtgouierg:
guys, which issue is incorrect cursor placement when there're unicode characters of different width?

this one #5431 or #4142 or #1153 ?

I'm having working fix - (available in my fork for testing) - but i haven't added the fix to the selection yet (selection should use same adjustment)

(if someone is familiar with Selection class enough to point me to where it calculates the coordinates, by the way - you could help me to do it sooner)

ewnjtgouierg:
i fixed this one (combining diacritics) for cursor positioning too, but ace still detects the end of line incorrectly, so if you just press right arrow you get stuck at EOL until you hit one for each diacritic lol

gotta fix that

is there anyone who's been working on ace for a while around? it could really make it all faster if i could consult with someone about what in ace is done where and how, instead of reverse-engineering everything

ewnjtgouierg:
@nightwing , thank you again for being willing to help

as you know, Cursor::getPixelPosition() just multiples column number on a fixed width
which doesn't work for wider characters like emojis

this i fixed (returning actual position instead)

[...]

Issue #4602, opened 2021-12-23
<U+FE0F> leads to cursor missalignment


Sivecano:
characters like <U+FE0F> get rendered with zero width with no indication of there being a character there. this miss-aligns the cursor.
it would probably be a good idea to treat characters like that (especially when alone) the same as <U+2000D>:
clearly marking them and giving them width.

Sivecano:
I have been able to consistently reproduce it in the kitchen sink on Linux by simply inserting the character.  I have also been able to reproduce it using both firefox and chrome in a windows 10 virtual machine.
the cursor misalignment still happens on both firefox and chrome. in the image below you can see the syntax-highlighting not recognising the word "function" because there is an <U+FE0F> between the some of the letters and the bracket highlights getting misaligned.
![ace_bug](https://user-images.githubusercontent.com/49620143/168578205-6bd3fe3d-12de-4e40-9be7-28702014d008.jpg)
this is a plain-text copy of the code:
```javascript
fu️nction foo(items, nada) {
    for (var i=0; i<items.length; i++) {
        alert(items[i] + "juhu\n");
    }	// Real Tab.
}
```

InspiredGuy:
I was not able to reproduce the unrendered character, but the bracket highlighting misalignment is reproducible on v1.21.0 in [kitchen sink](https://ace.c9.io/build/kitchen-sink.html). Chrome, macOS.
![image](https://github.com/ajaxorg/ace/assets/1381598/4a91feb4-601d-4c30-942b-266c45d6cc64)

Sivecano:
after testing in https://ace.c9.io/build/kitchen-sink.html this is still not fixed. Tested on firefox on linux.

PR #5922, opened 2026-02-22
Initial support for using range.getClientRects to measure text.

nightwing:
[Open kitchen-sink @ c7be8dd97af41183600195164d17e00f387c23f6](https://raw.githack.com/ajaxorg/ace/c7be8dd97af41183600195164d17e00f387c23f6/kitchen-sink.html)

[Open kitchen-sink @ 90b4eee66710148337aee06de5ad6e46df0b347f](https://raw.githack.com/ajaxorg/ace/90b4eee66710148337aee06de5ad6e46df0b347f/kitchen-sink.html)

[Open kitchen-sink @ 5a08e666570cde601c49a89c3cf7ee66d32ba8de](https://raw.githack.com/ajaxorg/ace/5a08e666570cde601c49a89c3cf7ee66d32ba8de/kitchen-sink.html)

[Open kitchen-sink @ bed04381bebaa0464636dbed3dbd1c95e6edc99e](https://raw.githack.com/ajaxorg/ace/bed04381bebaa0464636dbed3dbd1c95e6edc99e/kitchen-sink.html)

[Open kitchen-sink @ 2de603755db99236770c3d1d096aac6583f4e8bf](https://raw.githack.com/ajaxorg/ace/2de603755db99236770c3d1d096aac6583f4e8bf/kitchen-sink.html)

[Open kitchen-sink @ 6138c2d3378fc9b489a279145fc98be977233005](https://raw.githack.com/ajaxorg/ace/6138c2d3378fc9b489a279145fc98be977233005/kitchen-sink.html)

[Open kitchen-sink @ d7b2cdd7cb12149184fd7f0f0475cc2f846773c1](https://raw.githack.com/ajaxorg/ace/d7b2cdd7cb12149184fd7f0f0475cc2f846773c1/kitchen-sink.html)

[Open kitchen-sink @ a99a4870bb7d2ec8e7d16cba766645c75e0d713c](https://raw.githack.com/ajaxorg/ace/a99a4870bb7d2ec8e7d16cba766645c75e0d713c/kitchen-sink.html)

[Open kitchen-sink @ 60ce0b14fe5032b4fd5789871fa578a652d72fa5](https://raw.githack.com/ajaxorg/ace/60ce0b14fe5032b4fd5789871fa578a652d72fa5/kitchen-sink.html)

[Open kitchen-sink @ 33d1040444586512d70f0d4c28302e94879cf2da](https://raw.githack.com/ajaxorg/ace/33d1040444586512d70f0d4c28302e94879cf2da/kitchen-sink.html)

[Open kitchen-sink @ 0e9e8bec27cd3677fc50a77343426c4e7bd09a39](https://raw.githack.com/ajaxorg/ace/0e9e8bec27cd3677fc50a77343426c4e7bd09a39/kitchen-sink.html)

[Open kitchen-sink @ fca477cbaef5a4f1917382eeed2fd300ac928c60](https://raw.githack.com/ajaxorg/ace/fca477cbaef5a4f1917382eeed2fd300ac928c60/kitchen-sink.html)

[Open kitchen-sink @ c6c9f87a587d99fb015adf3c905339d16bc54ee0](https://raw.githack.com/ajaxorg/ace/c6c9f87a587d99fb015adf3c905339d16bc54ee0/kitchen-sink.html)

[Open kitchen-sink @ 32ae297abead042c6680563f46b8e10dcb6483ec](https://raw.githack.com/ajaxorg/ace/32ae297abead042c6680563f46b8e10dcb6483ec/kitchen-sink.html)

[Open kitchen-sink @ 40e5bdc846720749cb5392b96a1181744b87c3bb](https://raw.githack.com/ajaxorg/ace/40e5bdc846720749cb5392b96a1181744b87c3bb/kitchen-sink.html)

[Open kitchen-sink @ 8f9018cdcadd14972e6cd8d2c6aa0c6a1ba25aeb](https://raw.githack.com/ajaxorg/ace/8f9018cdcadd14972e6cd8d2c6aa0c6a1ba25aeb/kitchen-sink.html)

[Open kitchen-sink @ 8105fe2f6937b345a923c92a95c461dc3e51e11b](https://raw.githack.com/ajaxorg/ace/8105fe2f6937b345a923c92a95c461dc3e51e11b/kitchen-sink.html)

[Open kitchen-sink @ bd38fd5337efe5242de0ff6edaaa75fa75e8ec5c](https://raw.githack.com/ajaxorg/ace/bd38fd5337efe5242de0ff6edaaa75fa75e8ec5c/kitchen-sink.html)


src/layer/font_metrics.js
```diff
@@ -207,8 +209,200 @@ class FontMetrics {
...
+        var row = startScreenPos.row;
+        var textLayer = this.textLayer;
+        var lineElement = this.$findElementForScreenRow(row);
+
+        if (lineElement) {
+            try {
+                var p1 = this.$findColumnPosition(lineElement, startScreenPos.column);
+                var p2 = this.$findColumnPosition(lineElement, endScreenPos.column);
+
+                this.$scratchRange.setStart(p1.node, p1.offset);
```
xyos: p1 could be null right, will this fail in this case?

src/virtual_renderer.js
```diff
@@ -1644,9 +1645,11 @@ class VirtualRenderer {
 
         var offsetX = x + this.scrollLeft - canvasPos.left - this.$padding;
         var offset = offsetX / this.characterWidth;
-        var row = Math.floor((y + this.scrollTop - canvasPos.top) / this.lineHeight);
+        var row = (y + this.scrollTop - canvasPos.top) / this.lineHeight;
```
xyos: any reason to remove the floor on this?

src/layer/font_metrics.js
```diff
@@ -381,8 +443,6 @@ function mergeTouchingRects(rects) {
             if (
                 (m.left <= rect.left && rect.left <= m.right) ||
                 (m.left <= rect.right && rect.right <= m.right) ||
-                (m.left <= rect.right && rect.right <= m.right) || 
-                (m.left <= rect.left && rect.right <= m.right) ||
```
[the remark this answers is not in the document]
nightwing: yes `rect.left<=rect.right` so `rect.right<=m.right` is same as `rect.left<=m.right`

xyos (approved):
lgtm

src/layer/text.js
```diff
@@ -350,7 +350,7 @@ class Text {
 
     $renderToken(parent, screenColumn, token, value) {
         var self = this;
-        var re = /(\t)|( +)|([\x00-\x1f\x80-\xa0\xad\u1680\u180E\u2000-\u200f\u2028\u2029\u202F\u205F\uFEFF\uFFF9-\uFFFC\u2066\u2067\u2068\u202A\u202B\u202D\u202E\u202C\u2069\u2060\u2061\u2062\u2063\u2064\u206A\u206B\u206B\u206C\u206D\u206E\u206F]+)|(\u3000)|([\u1100-\u115F\u11A3-\u11A7\u11FA-\u11FF\u2329-\u232A\u2E80-\u2E99\u2E9B-\u2EF3\u2F00-\u2FD5\u2FF0-\u2FFB\u3001-\u303E\u3041-\u3096\u3099-\u30FF\u3105-\u312D\u3131-\u318E\u3190-\u31BA\u31C0-\u31E3\u31F0-\u321E\u3220-\u3247\u3250-\u32FE\u3300-\u4DBF\u4E00-\uA48C\uA490-\uA4C6\uA960-\uA97C\uAC00-\uD7A3\uD7B0-\uD7C6\uD7CB-\uD7FB\uF900-\uFAFF\uFE10-\uFE19\uFE30-\uFE52\uFE54-\uFE66\uFE68-\uFE6B\uFF01-\uFF60\uFFE0-\uFFE6]|[\uD800-\uDBFF][\uDC00-\uDFFF])/g;
+        var re = /(\t)|( +)|([\x00-\x1f\x80-\xa0\xad\u1680\u180E\u2000-\u200f\u2028\u2029\u202F\u205F\uFEFF\uFFF9-\uFFFC\u2066\u2067\u2068\u202A\u202B\u202D\u202E\u202C\u2069\u2060\u2061\u2062\u2063\u2064\u206A\u206B\u206B\u206C\u206D\u206E\u206F]+)|(\u3000+)/g;
```
xyos: This control-character class still includes U+200D (ZWJ) via the `\u2000-\u200f` range, so joined emoji like 👨‍👩‍👧‍👦 render chopped up by invalid-char dots. Splitting the range to `\u2000-\u200C\u200E\u200f` keeps ZWJ sequences intact and fixes #5813. (This is the one-character version of what I did in #5988.)

src/layer/font_metrics.js
```diff
@@ -193,23 +183,516 @@ class FontMetrics {
...
+                return screenColumn = screenColumn + index;
+            } else if (node.nodeType === Node.ELEMENT_NODE) {
+                var childNodes = node.childNodes;
+
+                for (var i = 0; i < childNodes.length; i++) {
+                    var child = childNodes[i];
+                    var rects = getRects(child);
+                    for (var j = 0; j < rects.length; j++) {
+                        let rect = rects[j];
+                        if (rect.left < x && x < rect.left + rect.width) {
```
xyos: Strict inequalities here mean a click landing exactly on the seam between two spans (e.g. with `showInvisibles` on) matches neither child rect, falls through, and accumulates the wrong column. `rect.left <= x && x <= rect.left + rect.width` fixes it. Also worth clamping to column 0 when x is left of the line start; right now that path can return a large column.

src/layer/font_metrics.js
```diff
@@ -193,23 +183,516 @@ class FontMetrics {
...
+                    if (hasCssTransform) {
+                        rect = self.recoverRect(tr, rect);
+                    }
+                    var d = Math.abs(x - rect.left);
+                    if (d < maxDistance) {
+                        index = j;
+                        maxDistance = d;
+                    }
+                }
+                if (blockCursor) {
```
xyos, line 456: `if (blockCursor) { // TODO }` is a no-op, so vim block-cursor mode gets round-to-nearest instead of floor semantics. Either implement or drop the branch and note the limitation.

src/layer/cursor.js
```diff
@@ -175,22 +175,23 @@ class Cursor {
...
         var pos = this.session.documentToScreenPosition(position);
-        var cursorLeft = this.$padding + (this.session.$bidiHandler.isBidiRow(pos.row, position.row)
-            ? this.session.$bidiHandler.getPosLeft(pos.column)
-            : pos.column * this.config.characterWidth);
+        var textWidth = this.config.fontMetrics.textWidth(pos.row, pos.column);
+        var cursorLeft = this.$padding + textWidth;
 
         var cursorTop = (pos.row - (onScreen ? this.config.firstRowScreen : 0)) *
             this.config.lineHeight;
+        var cursorWidth = this.config.fontMetrics.textWidth(pos.row, pos.column + 1) - textWidth;
```
xyos, line 192: If `textWidth(col+1)` and `textWidth(col)` measure equal (zero-width cluster tail, or a measurement fallback), `cursorWidth` is 0 and the cursor becomes invisible. Suggest `(... - textWidth) || this.config.characterWidth`.

src/bidihandler.js
```diff
@@ -51,12 +51,7 @@ class BidiHandler {
     isBidiRow(screenRow, docRow, splitIndex) {
         if (!this.seenBidi)
             return false;
-        if (screenRow !== this.currentRow) {
-            this.currentRow = screenRow;
-            this.updateRowLine(docRow, splitIndex);
-            this.updateBidiMap();
-        }
-        return this.bidiMap.bidiLevels;
+        return true;
```
xyos, line 54: `isBidiRow` now returns true for every row once the document contains any bidi character. One Hebrew word in a 10k-line file routes every marker on every line through `drawBidiSingleLineMarker` → `getRects` → DOM range measurement. A cheap per-line `bidiRE.test(line)` here would keep the fast path for the 99% of lines that aren't bidi.

src/selection.js
```diff
@@ -711,15 +711,14 @@ class Selection {
...
-                if (this.session.$bidiHandler.isBidiRow(screenPos.row, this.lead.row)) {
-                    offsetX = this.session.$bidiHandler.getPosLeft(screenPos.column);
-                    screenPos.column = Math.round(offsetX / this.session.$bidiHandler.charWidths[0]);
-                } else {
-                    offsetX = screenPos.column * this.session.$bidiHandler.charWidths[0];
-                }
+            if (useFontMetrics && this.$desiredColumn == null) {
+                offsetX = fontMetrics.textWidth(screenPos.row, screenPos.column);
+                if (isFinite(offsetX))
+                    this.$desiredColumn = offsetX / fontMetrics.config.characterWidth;
```
xyos, line 721: `$desiredColumn` now stores a fractional pixel-derived value (`offsetX / characterWidth`). When the target row isn't rendered, `$pixelToColumn` returns it unchanged and a non-integer column flows into `screenToDocumentPosition`. It works in the cases I tested, but rounding the fallback would keep the column contract integer-valued.

demo/kitchen-sink/inline_editor.js
```diff
@@ -21,7 +21,7 @@ require("ace/commands/default_commands").commands.push({
             return;
         }
         
-        var rowCount = 10;
+        var rowCount = 5.5;
```
xyos, line 24: `rowCount = 5.5` looks like a leftover from testing fractional row heights; intentional?

xyos (commented):
Nice work, this is awesome!

I built #5988 on an earlier snapshot of this branch, so below are a few things I hit while testing grapheme/bidi edge cases against real browser measurements. Most are small and cherry-pickable from #5988. Details inline.

Once this merges I'd like us to review and maybe do a follow-up PR picking up the rest of #5988: the grapheme-cluster model (`Intl.Segmenter`-based helpers in `lib/lang`, cluster-snapping `moveCursorTo`, cluster-aware `$getDisplayTokens`/`$computeWrapSplits`, `$alignTokensToClusters`). It touches the same code this PR rewrites, so rebasing it on top of merged master is cleaner than growing this PR's scope. With the current branch a wrap limit crossing 👨‍👩‍👧‍👦 still splits its 11 code units across screen rows, and `moveCursorTo` only snaps surrogate pairs, not ZWJ/combining-mark clusters; the follow-up addresses #460, #4142, #5431, #4602, #3753, #3866, #3617.

src/layer/font_metrics.js
```diff
@@ -193,23 +183,516 @@ class FontMetrics {
...
+            lineElement =  lineElement.children[Math.max(index, 0)];
         }
-        var u = sub(clientPos, a);
-        var f = solve(sub(m1, mul(h[0], u)), sub(m2, mul(h[1], u)), u);
-        return mul(L, f);
+        return lineElement;
     }
     
+    /**
+     * Calculates the width of the text up to a specific scrrenColumn on a given screen row.
```
xyos: nit: typo, `scrrenColumn` (also ends up in the generated `types/ace-modules.d.ts`)

src/virtual_renderer_test.js
```diff
@@ -60,62 +61,90 @@ module.exports = {
...
-        var pos1 = fontMetrics.transformCoordinates(null, [0, 200]);
-        assert.ok(pos1[0] - rects[2][0] < 10e-6);
-        assert.ok(pos1[1] - rects[2][1] < 10e-6);
-        editor.renderer.$loop._flush();
+        editor.container.style.transformOrigin = "0 0";
+        var H1 = -0.0007, H2 = -0.001;
+        var m0 = 0.7, m1 = 0.1, m2 = 0.3, m3 = 0.82;
+        var t1 = 100, t2 = 20;
+        function testTransform() {
+            fontMetrics.config.$transformData = null; //FIXME
```
xyos: nit: remove this comment

types/ace-modules.d.ts
```diff
@@ -4600,13 +4635,12 @@ declare module "ace-code/src/edit_session" {
...
          * @param {Number} screenRow The screen row to check
          * @param {Number} screenColumn The screen column to check
-         * @param {Number} [offsetX] screen character x-offset [optional]
          *
          * @returns {Point} The object returned has two properties: `row` and `column`.
          *
          * @related EditSession.documentToScreenPosition
          **/
-        screenToDocumentPosition(screenRow: number, screenColumn: number, offsetX?: number): Point;
+        screenToDocumentPosition(screenRow: number, screenColumn: number): Point;
```
xyos, line 4643: This is a public API break: `screenToDocumentPosition` loses its documented `offsetX` param, and `BidiHandler` loses `getPosLeft`/`getSelections`/`offsetToCol`/`updateCharacterWidths`. Embedders calling these will break at runtime, not just in types. Probably fine given the old bidi pixel model is gone, but worth a changelog note and a decision on whether this ships as a minor or major bump.

src/bidihandler.js
nightwing: The problem is that when calling isBidiRow we only have screenLine, so finding the line is quite expensive. drawBidiSingleLineMarker on the other hand is not very slow, so we end up with similar performance in both cases.

demo/kitchen-sink/inline_editor.js
secret-account: fractional row heights were intentional, without that all linewidgets are forced to have heights in multtiples of line height, and our api doesn't enforce that.

src/selection.js
secret-account: I think we should allow fractional screenColumn for reasons similar to screenRow, also because fractional screenColumns were allowed in the old implementation of varchar https://github.com/ajaxorg/ace/blob/master/src/edit_session.js#L2507C19-L2507C34

src/layer/cursor.js
secret-account: good point, done

src/layer/font_metrics.js
secret-account: done

PR #5980, opened 2026-06-13
Add onlineminitools.com

prajwal89:
Added onlineminitools.com to the Real World Users list, as we extensively use Ace Editor across our tools. 
Thank you for your work!

[Open kitchen-sink @ 497df1aec147640ca33dc971e6ffc26714ad8f63](https://raw.githack.com/prajwal89/ace/497df1aec147640ca33dc971e6ffc26714ad8f63/kitchen-sink.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1224,6 +1224,11 @@ if (match) {
                                     <a href="https://chartbrew.com/">Chartbrew</a>
                                 </li>
 					    
+                                <li>
+                                    <img lazy-src="https://onlineminitools.com/images/logo.png"
+                                        style="left: 10px; top: -4px; width:80px" />
+                                    <a href="https://onlineminitools.com">OnlineMiniTools</a>
+                                </li>
                                 <li id="add_your_site">
                                     <p>+</p>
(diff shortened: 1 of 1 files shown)


rahulkumarsingh73690:
The website is not accessible

prajwal89:
> The website is not accessible

I checked the website from multiple locations, and it's working fine on my end. It may be a temporary issue on your side.

That said, I've also switched the nameservers to Cloudflare, so if the issue is related to DNS propagation or routing, it should be resolved shortly.

xyos:
same for me, site is not working

prajwal89:
Thanks for checking and reporting it.

I found the root cause; it turned out to be a rather sneaky SSL configuration issue that was only affecting the homepage from some US locations, which is why it was difficult to reproduce initially. I've fixed it now.

I also verified the fix using:
- https://semonto.com/tools/website-reachability-check?test=0ec4a168-34d6-4bf6-8963-f1d9c672c0bb
- https://geotargetly.com/geo-browse

Both confirm that the homepage is now loading correctly from the previously affected regions.

Could you please try again when you have a moment? Thanks!

Issue #5993, opened 2026-07-19
ace.js: search highlighting makes the editor unresponsive on large folded document


pageldev:
### Describe the bug

Search highlighting can make the editor very unresponsive when searching a large, heavily folded document.

### Expected Behavior

Editor should remain responsive independent of folded sections

### Current Behavior

Editor becomes unresponsive 

### Reproduction Steps

1. Open: https://sekedus.github.io/ace/kitchen-sink.html
2. Paste the contents of test.xodr.xml
3. Fold a lot of high level `<road>` elements (like 15-20)
4. Search for "Road" and scroll, the editor will be very unresponsive 

[test.xodr.xml](https://github.com/user-attachments/files/30163372/test.xodr.xml)

### Ace Version / Browser / OS / Keyboard layout

v1.44.0 / Safari / Mac OS 26 / DE

PR #5994, opened 2026-07-19
fix: avoid repeated document reads during search highlighting

pageldev:
*fixes Issue #5993*

Search highlighting became slow in large, folded documents because `session.getValue()` was called for every row being processed (which can be a lot if folded at high level). `session.getValue()` rebuilds the full document string causing a lot of repeated work. This was introduced in https://github.com/ajaxorg/ace/commit/06d51b9164dc9dc39fbd6df26b7132f4b9580a4c

*Description of changes:*
- Read the document value once per highlight update.
- Reuse its length when checking whether the highlight cache is stale.

[Open kitchen-sink @ 2af47ee796d23e1ad7cfc224a7e0cb512bfa8865](https://raw.githack.com/pageldev/ace/2af47ee796d23e1ad7cfc224a7e0cb512bfa8865/kitchen-sink.html)

### DIFF
--- a/src/search_highlight.js
+++ b/src/search_highlight.js
@@ -39,10 +39,12 @@ class SearchHighlight {
         var renderedMarkerRanges = {};
         var _search = session.$editor && session.$editor.$search;
         var mtSearch = _search && _search.$isMultilineSearch(session.$editor.getLastSearchOptions());
+        var docLen = session.getValue().length;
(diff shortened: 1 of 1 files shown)


nightwing:
Thanks, we probably could completely eliminate docLen by using change listener or undoManager $rev. because the current implementation was not only slow but also incorrect as changes not changing overal length have no effect.

PR #6007, opened 2026-09-13
fix: support a leading lookbehind in rules with token arrays

### LINKED ISSUE #6001: (extending syntax highlighting): (use of look behind breaks token array)
### Describe the bug

Using the `https://ace.c9.io/tool/mode_creator.html` tool, please consider the **TestRules** mode in the reproduction part of the Issue.

When the regex in the line labelled 'problematic line' is as listed below, each of the captured groups is mapped correctly to the array of tokens in the line labelled 'token array that breaks', which can be verified by hovering over each component of one of the 'metaoptions' (eg `:second-option<<<<<second thing>>>>>`)

This seems to be a bug as look ahead/behind assertions do work, but they just screw up the (documented) array mapping.  (I have not investigated whether the sub-match groups can be accessed by passing the match to a subroutine. In any case, there is either a bug here, maybe some undocumented syntax to extract the subgroups).

### Expected Behavior

The regex in the line labelled 'problematic line' should generate a match object with an array of subgroups that can be mapped into the array defined in `token`.

### Current Behavior

Adding a look-behind assertion allocates the matched text to the first of element of the array.

Prepend the first group with a look-behind assertion, the whole regex match fails.

For example,
```
regex: "(?<=\\s)(\\:[a-zA-Z][a-zA-Z0-9_-]+)([<({]+)([^>)}]+)([>)}]+)", //problematic line
\\           !!!!!!!!!!!!
\\ it doesnt matter whether the look-behind is inside or outside the first group bracket
```
breaks the regex to token mapper.

If the associated token line is changed to either of the forms below, the 'metaoption' **is** matched but only the token `"nop"` is allocated to the whole matched segment.
```
token: ["nop","s_metaoption", "lbrk", "s_option_cont", "rbrk"], // token array  that breaks
// or where the match is mapped to *s-metaoption*
//token: "s_metaoption", // token array  that breaks
```

### Reproduction Steps

Below is the working version of the test mode without the look behind assertion in the problematic line,
and here is a line of data to apply the mode to:
```
=document vvv :one-option<first> :second-option<<<<<second thing>>>>>
```
working code
```
"use strict";

var oop = require("../lib/oop");
var TextHighlightRules = require("./text_highlight_rules").TextHighlightRules;

var TestRules = function () {
  
  var metaoption_c = { 
    token: ["s_metaoption", "lbrk", "s_option_cont", "rbrk"], // token array  that breaks
    regex: "(\\:[a-zA-Z][a-zA-Z0-9_-]+)([<({]+)([^>)}]+)([>)}]+)", //problematic line
    next: "metaoptionLine"
  };
  
  this.$rules = {
		"start" : [
		    {
		        token: "s_document",
		        regex: "=document\\b",
		        next: "metaoptionLine"
		    }
	    ],
      "metaoptionLine" : [
          metaoption_c,
          {
              token: "invalid.illegal",
              regex: "(?<=\\s)[^\\s:]\\S+(?=\\s)",
              next: "metaoptionLine"
          },
          {
              token: "text",
              regex: "(\\s*)$",
              next: "start"
          },
      ]
  }
};

oop.inherits(TestRules, TextHighlightRules);

exports.TestRules = TestRules;
```

### Possible Solution

The match object with a look-behind assertion is more complex than a simple array. 

There is no *documented* method to access sub-matches

### Additional Information/Context

_No response_

### Ace Version / Browser / OS / Keyboard layout

Accessing https://ace.c9.io/tool/mode_creator.html using a Firefox browser.

lbesecker195:
*Issue #, if available:* Fixes #6001

*Description of changes:*

When a rule has more than one capturing group, the tokenizer splits the matched text with a second regexp built by `createSplitterRegexp`. That regexp runs on the matched text alone. A lookbehind at the start of the rule regexp has no text before it to check, so the split failed:

- With a token array, `$arrayTokens` got `null` and returned the whole match as a single `text` token. This is the behavior in #6001.
- With a token function, `$applyToken` threw `Cannot read properties of null (reading 'slice')`.

The full rule regexp has already checked the lookbehind against the line, so the splitter can drop it. `createSplitterRegexp` already does the same for a trailing lookahead. The new code removes `(?<=...)` and `(?<!...)` groups when nothing but `(` or `(?:` comes before them. That covers both forms from the issue, with the lookbehind inside or outside the first group. Escapes, character classes, and nested groups inside the lookbehind are skipped when finding its closing paren. A lookbehind later in the regexp is kept, because the text it checks is part of the match. Named groups such as `(?<name>...)` are left alone.

Tests in `src/tokenizer_test.js`:

- `createSplitterRegexp removes leading lookbehind` covers regexp sources for the cases above, plus a combined leading lookbehind and trailing lookahead.
- `token array with a leading lookbehind` uses the rule regexp and the sample line from #6001, and a token function with a leading lookbehind.

## Verification

| | `npx mocha src/tokenizer_test.js` | `npm test` |
|---|---|---|
| Before the `src/tokenizer.js` change | 3 passing, 2 failing (the new tests) | 1485 passing, 3 pending, 2 failing |
| After | 5 passing | 1487 passing, 3 pending, 2 failing |

The 2 failures in `npm test` are the same with and without this change: `test: multiselect` in `mouse_handler_test.js` and ``test vim_<C-c>copy`` in `vim_test.js`. I ran on macOS with Node, and they look environment related.

`npx eslint src/tokenizer.js src/tokenizer_test.js` and `npm run typecheck` pass. I did not test in the mode creator page.

Pull Request Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[Open kitchen-sink @ 3c1176684909a26f2a48cd0d00a4412a48148845](https://raw.githack.com/lbesecker195/ace/3c1176684909a26f2a48cd0d00a4412a48148845/kitchen-sink.html)

### DIFF
--- a/src/tokenizer.js
+++ b/src/tokenizer.js
@@ -203,6 +203,36 @@ class Tokenizer {
             if (lastCapture.end != null && /^\)*$/.test(src.substr(lastCapture.end)))
                 src = src.substring(0, lastCapture.start) + src.substr(lastCapture.end);
         }
+
+        // The splitter only sees the matched text, so a lookbehind at the start
+        // has no preceding text to check. It was already checked by the full
+        // regexp, so remove it, like the trailing lookahead above.
+        var leadingLookbehind;
+        while ((leadingLookbehind = /^(?:\((?:\?:)?)*\(\?<[=!]/.exec(src))) {
+            var start = leadingLookbehind[0].length - 4;
+            var end = -1;
+            var depth = 0;
+            var inClass = false;
+            var re = /\\.|[\[\]()]/g;
+            re.lastIndex = start;
+            var m;
+            while ((m = re.exec(src))) {
+                var ch = m[0];
+                if (ch.length > 1) continue;
+                if (inClass) {
+                    inClass = ch != "]";
+                } else if (ch == "[") {
+                    inClass = true;
+                } else if (ch == "(") {
+                    depth++;
+                } else if (ch == ")" && --depth == 0) {
+                    end = m.index + 1;
+                    break;
+                }
+            }
+            if (end == -1) break;
+            src = src.substring(0, start) + src.substr(end);
+        }
         
         // this is needed for regexps that can match in multiple ways
         if (src.charAt(0) != "^") src = "^" + src;
--- a/src/tokenizer_test.js
+++ b/src/tokenizer_test.js
@@ -20,6 +20,71 @@ module.exports = {
         assert.equal(re.source, "^[(?=)](\\?=t)$");
     },
 
+    "test: createSplitterRegexp removes leading lookbehind" : function() {
+        var t = new Tokenizer({});
+        var re = t.createSplitterRegexp("(?<=\\s)(a)(b)");
+        assert.equal(re.source, "^(a)(b)$");
+        var re = t.createSplitterRegexp("((?<=\\s)a)(b)");
+        assert.equal(re.source, "^(a)(b)$");
+        var re = t.createSplitterRegexp("(?<![x)(])(a)(b)");
+        assert.equal(re.source, "^(a)(b)$");
+        var re = t.createSplitterRegexp("(?:(?<=(x|\\)))a)(b)");
+        assert.equal(re.source, "^(?:a)(b)$");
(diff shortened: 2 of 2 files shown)


nightwing:
Thank you!

